PR #3, opened 2018-12-11
Update electron_counting.py

HamishGBrown:
Fixed two bugs:

Random sampling of frames for histogram fitting of dark reference (line 78) sometimes overshot array bounds.

Center of mass was being performed over a 2x2 local region (line 194-196) and this has been corrected to a 3x3 region - typical Python array bounds gotcha


HamishGBrown:
Hi Ben,

Would you mind pulling these changes to the electron counting routine in preparation for our skype discussion?

PR #7, opened 2019-04-11
Updates for install

mlhenderson:
1. Fixes version file not found after install
- Moved version.py under py4DSTEM
- updated setup.py exec to point to new location for version.py module
- updated py4DSTEM/file/log.py to import __version__ from version.py module instead of exec

2. Adds required packages to setup.py under install_requires, with compatible versions.


bsavitzky:
Awesome, looks great, thanks for fixing this Matt

PR #15, opened 2019-05-10
Relativity

sezelt:
Adds new `load='relativity'` option to `py4DSTEM.file.readwrite.read(...)` which opens an `*.mrc` file as a memory map. Adds `relativity.py` to `process.preprocess` which contains functions for:
* aligning subframes of a stack (`subframe_align`)
* `slice_mrc_stack`: converting the (3D) `*.mrc` stack stack into a `DataCube`. This operation requires a "scratch" file where a numpy memory map is buffered. Presumably, one would just save a py4DSTEM file to retain the stack and all the metadata after this, and get rid of/overwrite the scratch file. 

This adds dependency on `mrcfile` (which has to be `pip install`ed because it's not in conda) and uses the `FloatProgress` doodad from `ipywidgets` for a progress bar (which in my experience is quite sturdy). 

Also, I'm attaching a notebook with a usage example (though not the dataset, since it's several GB)
[usage example](https://github.com/py4dstem/py4DSTEM/files/3168037/module.test.ipynb.zip)

### DIFF
--- a/py4DSTEM/file/readwrite/read.py
+++ b/py4DSTEM/file/readwrite/read.py
@@ -19,7 +19,6 @@
 import hyperspy.api as hs
 from .dm import dmReader
 from .empad import read_empad
-from ...process.preprocess import read_mrc_file
 from .filebrowser import FileBrowser, is_py4DSTEM_file
 from ..datastructure import DataCube
 from ..datastructure import Metadata
@@ -76,8 +75,9 @@ def read(filename, load=None):
             print("Reading an EMPAD file...")
             output = read_empad_file(filename)
         elif load == 'relativity':
+            import mrcfile
             print("Reading an IDES Relativity MRC file...")
-            output = read_mrc_file(filename)
+            output = mrcfile.mmap(filename,mode='r')
         else:
             print("Error: unknown value for parameter 'load' = {}. Returning None. See the read docstring for more info.".format(load))
             output = None
--- a/py4DSTEM/file/readwrite/read.py
+++ b/py4DSTEM/file/readwrite/read.py
@@ -19,6 +19,7 @@
 import hyperspy.api as hs
 from .dm import dmReader
 from .empad import read_empad
+from ...process.preprocess import read_mrc_file
 from .filebrowser import FileBrowser, is_py4DSTEM_file
 from ..datastructure import DataCube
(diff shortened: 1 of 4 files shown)


bsavitzky:
This is great, thanks Steven!

I made the following edits:
-documentation edits. Mostly fleshing out your documentation in a few places, as well as some reformatting for consistency with other py4DSTEM docstrings.
-get_shifts/get_shifts_hybrid.  Having both was redundant, so I just replaced the old get_shifts function with your improved version, as functionality is the same if the corrPower is left at its default of 1.

There are a few other edits that I think would help this module, including one very minor and one more major edit that I think are needed, plus a couple more optional edits / suggestions:
- major: add a demo notebook - our model so far has been including demos for all py4DSTEM functionality in notebooks, mostly in the ./notebooks/module_demos/ directory.  I would add something like ./notebooks/module_demos/preprocess/relativity.ipynb - I know you attached a demo notebook in your pull request message, so perhaps it's already made!  The plan is for this to eventually turn into documentation pages in a github.io site, so clear descriptions of what's happening in comments and/or markdown cells are bonuses.
- minor: add descriptions of the outputs of the subframe_align function (relativity.py, lines 74-76)
- suggestion 1: relativity.py includes some plotting within some of your function calls - as a general rule I've been trying to keep plotting / visualization out of module function calls, which keeps things cleaner andeasier to navigate.  I would suggest removing this stuff from your functions, and instead perhaps adding a boolean flag as a kwarg to your function which, if True, tells the function to return all the pieces you need to make the plots you want.  Then do the plotting outside of the function call (in your jupyter notebook or script or whatever you like).
- suggestion 2: I don't remember suggest 2 anymore
- suggestion 3: Oh right, this is what it was: consider adding assert statements to the beginnings of your functions, to make sure things don't break (and/or when they do, that you'll have an easier time figuring out why).
- suggestion 4: consider limiting frequency of stream-of-consciousness ramblings in github messages, or, minimally, no longer referring to them as 'suggestions' (for internal use only)

Annnnnnyway, this is all in my fork under a branch called stevens_pr.  Would you like to pull that, take a look, keep/remove my edits/suggestions as you see fit, and then we can loop back on merging into the central repo?

Thanks again for the new module, it looks excellent!
Ben

PR #20, opened 2019-05-16
Add multicorr subpixel fitting to Bragg disk detection

sezelt:
Adds the multicorr DFT-upsampling algorithm to the `find_bragg_disk...` family of functions. 

**This changes the behavior of the `subpixel` option in the `find_bragg_disk` family of functions to now accept one of three string arguments instead of a boolean** ('none', 'poly', and 'multicorr'). 

Subpixel registration can go two ways:
* Take an FFT, zero-pad the array, and inverse transform. This is simple to code and involves a _fast operation_ acting on _(very) large data_. 
* Take and FFT and compute its inverse by matrix-multiplication DFT. Using a modified transformation matrix, we can transform only a partial area of the inverse DFT. This is nasty to code, and involves a _slow operation_ acting on _small data_.

For large upsample factors, and when we only _need_ to upsample a very small neighborhood, the matrix-multiply method is faster and more efficient. See, for example, [this paper](https://doi.org/10.1364/OE.15.015935) or [this paper](https://doi.org/10.1364/OL.33.000156).

In previous implementations (such as the MATLAB one) we do this in two steps: first FFT-zero-pad upsample the whole correlogram to half-pixel precision, then use the DFT upsampling in a 1.5px region around the identified peaks. Here, I have taken a slightly different approach: I use the parabolic subpixel fitting in `get_maxima_2D` and round to half-pixel values, then do the DFT upsample. This is partly out of laziness, partly because it means not having to normalize the thresholding parameters between the subpixel methods, and partly because I think it might be faster. (FYI multicorr is approx 3 times slower than parabolic on my machine, and parabolic is basically free).

I haven't added `multicorr` as an option in `get_maxima_2D`. I can if we want to, I just thought it would be potentially more disruptive since so many other things call that function.


bsavitzky:
Resolved a small merge conflict.  Thanks for getting the multicorr code up and running, Steven!

PR #26, opened 2019-06-07
Typo in remove.py

sezelt:
This seems to be a typo... Though I'm also confused about the structure of these three functions because they look redundant?

### DIFF
--- a/py4DSTEM/file/io/remove.py
+++ b/py4DSTEM/file/io/remove.py
@@ -71,7 +71,7 @@ def remove_dataobject(index, filepath):
     assert isinstance(index, (int,np.integer))
 
     # Remove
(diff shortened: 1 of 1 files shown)


bsavitzky:
Thanks for catching this typo, Steve.  And you're right, the functions are, indeed, redundant - remove_from_index_list() and remove() are all we actually need.  I'll just take out the redundant function.  We certainly could make this one function instead of two, but I prefer the structure of having one function that does the work (remove_from_index_list) and one that parses the arguments and which users actually call (remove), as it separates out the important behavior (removing objects from files) and the API/io details (i.e. this function is overloaded, functioning slightly differently on an int or on a list of ints).  I've used this architecture in various places in the code, and I'm still figuring out how to do it best, hence the redundancy here.

PR #27, opened 2019-06-07
Add ideal lattice, PointListArray indexing, bragg vector intensity virtual imaging

sezelt:
Adds three functions to `latticevectors`:

`generate_lattice` returns a PointList with linear combinations of `u` and `v` vectors, to use in lattice vector fitting. Especially in polycrystalline data with imperfect classification (or overlapping grains) thresholding of the Bragg vector map will give peaks that don't correspond to the desired lattice. This function guarantees that only the desired lattice, with consistent indexing, is used for strain, etc.

`index_point_list_array` adds `'h'` and `'k'` coordinates to a PointListArray and indexes the peaks based on the input indexed PointList. Accepts a mask so can be used separately on classified regions.

`bragg_vector_intensity_map_by_index` returns a map of the intensity of the `(h,k)` indexed spots from an indexed PointListArray. Used to generate VDF images for a given index (though the intensity is the correlation intensity, not a background-corrected diffraction image intensity 🤷‍♀ ).

I haven't implemented actually making a mosaic of VDF images though I think that these utilities make it easier for someone/us to implement it on a case-by-case basis.


py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+        x -= x0
+        h_max = np.max(np.ceil(np.abs((x/ux,y/uy))))
+        k_max = np.max(np.ceil(np.abs((x/vx,y/vy))))
+    
+    (hlist,klist) = np.meshgrid(np.arange(-h_max,h_max+1),np.arange(-k_max,k_max+1))
+
+    M_ideal = np.vstack((hlist.ravel(),klist.ravel())).T
+    ideal_peaks = np.matmul(M_ideal,beta)
+
+    ideal_data = np.core.records.fromarrays(np.array([ideal_peaks[:,0],ideal_peaks[:,1],M_ideal[:,0],M_ideal[:,1]]),\
```
bsavitzky: This syntax works, but I find calling np.core.records.fromarrays to make np structured arrays somewhat inelegant.  I'd recommend instead using:

'''
coords = [('qx',float),('qy',float),('h',int),('k',int)]
ideal_data = np.zeros(len(ideal_peaks[:,0]), dtype=coords)
ideal_data['qx'] = ideal_peaks[:,0]
ideal_data['qy'] = ideal_peaks[:,1]
ideal_data['h'] = M_ideal[:,0]
ideal_data['k'] = M_ideal[:,1]
'''

Then you can just proceed from line 102.  It's much more verbose, but I also find it more clear, which I will choose over succinctness every time, and avoids digging up functions from deep in the guts of numpy.

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+
+def index_point_list_array(pointlistarray, lattice, maxPeakSpacing, mask=None):
+    """
+    Indexes the Bragg peaks in a PointListArray based on the indexed points in
+    lattice, which are placed in 'h' and 'k' coordinates in the PointLists in the array.
+    If `mask` is specified, only the locations where mask is True are indexed.
+
+    Accepts:
+        pointlistarray          (PointListArray) the braggpeaks pointlist to index
+        lattice           (PointList) a PointList containing with indexed peaks,
```
bsavitzky: Missing a tab

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+    # trim peaks outside the image
+    deletePeaks = (ideal_lattice.data['qx'] > Q_Nx) | \
+            (ideal_lattice.data['qx'] < 0) | \
+            (ideal_lattice.data['qy'] > Q_Ny) | \
+            (ideal_lattice.data['qy'] < 0)
+    ideal_lattice.remove_points(deletePeaks)
+
+    return ideal_lattice
+
+def index_point_list_array(pointlistarray, lattice, maxPeakSpacing, mask=None):
```
bsavitzky: Recommend changing to 

def index_pointlistarray(...

for consistency.

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+
+    assert isinstance(pointlistarray,PointListArray)
+    assert np.all([name in pointlistarray.dtype.names for name in ('qx','qy','intensity')])
+    assert isinstance(lattice, PointList)
+    assert np.all([name in lattice.dtype.names for name in ('qx','qy','h','k')])
+
+    if mask is None:
+        mask = np.ones(pointlistarray.shape,dtype=bool)
+
+    assert np.all(mask.shape == pointlistarray.shape), 'mask must have same shape as pointlistarray'
```
bsavitzky: Remove np.all (mask.shape and pointlistarray.shape are tuples, so the == statement will always be just True or False)

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+
+                for i in range(pl.length):
+                    r2 = (pl.data['qx'][i]-lattice.data['qx'])**2 + \
+                         (pl.data['qy'][i]-lattice.data['qy'])**2
+                    ind = np.argmin(r2)
+                    if r2[ind] <= maxPeakSpacing**2:
+                        pl.data['h'][i] = lattice.data['h'][ind]
+                        pl.data['k'][i] = lattice.data['k'][ind]
+                    else:
+                        pl.data['h'][i] = 8675309 # turns out int's can be NaN
```
bsavitzky: Hmm.  I'm not crazy about this approach.  Perhaps we should define an additional coord which contains a boolean mask, specifying where the index values are meaningful.  Unfortunately, I don't think we want to use np.ma here, as but simply adding a mask coord could do the trick.

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+    """
+    Returns a correlation intensity map for an indexed (h,k) Bragg vector
+    Used to obtain a darkfield image corresponding to the (h,k) reflection
+    or a bightifled image when h=k=0
+
+    Accepts:
+        pointlistarray      (PointListArray) a PointListArray with (at least) the coordinates
+                                'h', 'k', and 'intensity'
+        h, k                (int) indices for the reflection to generate an intensity map from 
+        symmetric           (bool) if set to true, returns sum of intensity of (h,k), (-h,k),
```
bsavitzky: Ah, nice :)

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
...
+                        pl.data['h'][i] = lattice.data['h'][ind]
+                        pl.data['k'][i] = lattice.data['k'][ind]
+                    else:
+                        pl.data['h'][i] = 8675309 # turns out int's can be NaN
+                        pl.data['k'][i] = 8675309
+
+    return pointlistarray
+
+
+def bragg_vector_intensity_map_by_index(pointlistarray,h,k, symmetric=False):
```
bsavitzky: We can add this later, but the way I've implemented this previously is to calculate all the dark field images at once, for efficiency reasons: at the cost of a little extra memory use - i.e. defining arrays for all the h,k combos upfront - you can populate all of these in a single loop over scan positions.

py4DSTEM/process/latticevectors/index.py
```diff
@@ -53,3 +53,158 @@ def index_bragg_directions(x0, y0, ux, uy, vx, vy, bragg_x, bragg_y):
 
     return h,k, bragg_directions
 
+def generate_lattice(ux,uy,vx,vy,x0,y0,Q_Nx,Q_Ny,h_max=None,k_max=None):
+    """
+    Returns a full reciprocal lattice stretching to the limits of the diffraction pattern
+    by making linear combinations of the lattice vectors up to (±h_max,±k_max).
+
+    This can be useful when (for whatever reason, such as polycrystallininty or weird intensities)
```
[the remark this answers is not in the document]
sezelt: fixed

py4DSTEM/process/latticevectors/index.py
sezelt: fixed

py4DSTEM/process/latticevectors/index.py
sezelt: fixed

py4DSTEM/process/latticevectors/index.py
sezelt: fixed
(I forgot the convention 😛)

py4DSTEM/process/latticevectors/index.py
sezelt: 👍

py4DSTEM/process/latticevectors/index.py
sezelt: now adds `'index_mask'` coordinate

py4DSTEM/process/latticevectors/index.py
sezelt: 😇

py4DSTEM/process/latticevectors/index.py
sezelt: yes let's save for later...

sezelt:
implemented your suggestions

PR #35, opened 2019-07-02
Add support for Gatan K2 files via a custom memory map

sezelt:
This provides an interface to a set of Gatan binary files. The resulting object is not actually a numpy memmap, but can be sliced like a numpy array: `dc.data4D[Rx,Ry,:,:]` and all slicing operations return a numpy `ndarray`. Loaded by calling `py4DSTEM.file.io.read(filename,load='gatan_bin')`. Any of the raw data files or just the folder can be passed as the filename.

By default, uses the hidden extra pixels in the data to subtract readout noise. The user can also specify their own dark reference instead. The readout subtracted diffraction pattern is returned whenever you slice the data, so you don't need to make a new datacube to apply the readout noise!

The whole object is not compatible with most numpy functions, so generating an average image or average DP can only be done by iterating manually. Perhaps these should be added as methods of the object?


bsavitzky:
Awesome new functionality, thanks Steven

PR #39, opened 2019-07-05
fixes to GUI setup, now works on macOS

sezelt:
changed the order of some Qt functions, GUI now works properly on macOS


bsavitzky:
Funny that the GUI was working on my mac but not yours.  Silly qt, what are you even thinking?  Thanks, Steven.

PR #41, opened 2019-07-09
v0.5 file IO; changed Dataslice/cube classes to use data field

lerandc:
Added functionality to read simulation files (mainly prismatic) as emd_group_type 2 as py4DSTEM v0.5; Dataslices and datacubes now use data instead of data2D and data4D as their data objects; changed all instances of data2D and data4D to data to prevent breaking of other things


bsavitzky:
Fanstastic.  Thanks Luis!!

PR #43, opened 2019-07-09
Fix type error in gatan K2

sezelt:
Fix a mismatched dtype in gatank2.py


bsavitzky:
Just to confirm - we're casting to int16 now - this won't cause overflow errors?

PR #50, opened 2019-07-17
Change: viewer now checks to see if file exists

HamishGBrown:
Hi Ben,

Just added a function so that the viewer checks to see if the path you've placed in `self.settings.data_filename.val` actually exists rather than attempting to opening and crashing. Other changes I think you've already implemented.

Hamish


bsavitzky:
Huzzah!  Thanks Hamish.

sezelt:
Ben, just wanted to confirm you meant to merge this into master (this seems to have caused master to diverge and be ahead of dev)...

HamishGBrown:
Whoops my bad, I've been working on the master branch in my fork. Switched to dev branch.

bsavitzky:
Ah, but I should've double checked which branch I was pulling into.  Alas!  Good catch, Steven.  I've just reverted this change in the master branch - Hamish, can you submit a new PR into dev?

PR #52, opened 2019-07-18
Gui arrowkey

sezelt:
I showed Karen the GUI and she requested this feature... So here it is! (Also she thinks the GUI is awesome!)

Qt handles keyboard events in... interesting ways? So I ended up making a radio button for where you want control to go. Since once you attach keyboard events to the widget it really holds on to it, I have set it up so if you press anything other than an arrow key it leaves arrowkey mode. Otherwise even putting the cursor in other places wouldn't wrest control and people might get confused/angry...

### DIFF
--- a/py4DSTEM/gui/dialogs.py
+++ b/py4DSTEM/gui/dialogs.py
@@ -398,12 +398,37 @@ class VirtualDetectorsWidget(QtWidgets.QWidget):
         self.buttonGroup_DetectorMode.setId(self.radioButton_CoMX, 3)
         self.buttonGroup_DetectorMode.setId(self.radioButton_CoMY, 4)
 
+        # Arrowkey Control
+        arrowkey_widget = QtWidgets.QWidget()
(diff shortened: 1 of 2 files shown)


HamishGBrown:
This would be amazing, Good request 👍

tcpekin:
This is absolutely my most used tool in the Matlab code I've written... this would be extremely nice. You could also see how hyperspy does it - their signal2D.plot() implementation is very nice as well, where it shows both a virtual DF and corresponding diffraction pattern, and you can either use arrow keys or click the VDF to move around.

sezelt:
@tcpekin That's a good idea, hyperspy does handle this well. I just looked into how they do it, and I can't seem to get the same behavior to work in py4DSTEM. There's something I'm not understanding about how Qt dispatches key events and how to control it, so I end up either intercepting every keypress all of the time or none of them ever...

tcpekin:
@sezelt I avoid Qt like the plague.. I'm happy you figured out at least how to intercept something.. Nice work.

PR #62, opened 2019-07-25
Add GUI for strain mapping

sezelt:
**Adds GUI for mapping strain!**
Everything *seems* to work, and I've tested it on various data, on Mac and Windows, and I've been able to squash the bugs that have come up. 

Most of the changes are in the new files added in `gui/strain` but there are little tweaks in other places that add the strain window, improve robustness, and improve integration with Qt.

There is a short readme in the strain folder that has the gist of how to use it.

One "hacky" thing in there is that pyqtgraph does not have a way of specifying an alpha mask without giving up the very nice histogram/colormap interactivity. So I create a custom child of their `ImageItem` that throws away the alpha channel from the colormap (which is always 1 for the LUTs we use) and replace it with a separately specified mask.


bsavitzky:
STRAIN MAP GUI TO THE FACE

PR #63, opened 2019-08-09
RDF module

lerandc:
Added methods for performing polar-elliptical transforms with a two-sided gaussian fit function a-la Colin, as well as a class for single atom scattering factors and a module with functions for calculating radial distribution functions. A demo notebook for the RDF module was also added, which documents the functions in the module and how to use them on an amorphous Ta data set.


bsavitzky:
Ok, finally getting to this PR!  Thanks for all this work Luis, this is fantastic.

Edits for the future that this module will need:
- Cleaner documentation, including descriptions of function inputs/outputs, and docstrings describing what all funcitons do.  This is there in some places but lacking in a few key places - e.g. during the final RDF calculation, in get_phi, etc.
- A few key TODO items can get filled in at some point, e.g. fitting the overall scaling factor to the atomic scattering factors.

PR #68, opened 2019-08-28
GatanK2: Add some support for numpy functions 

sezelt:
This adds limited duck-typed support for numpy functions to the K2DataArray memory map.

The following are now valid for data read with the `load='gatan_bin'` option:

```python
avgDP = np.mean(dc.data,axis=(0,1))
avgImg = np.mean(dc.data,axis=(2,3))

sumDP = np.sum(dc.data,axis=(0,1))
sumImg = np.sum(dc.data,axis=(2,3))

maxDP = np.max(dc.data,axis=(0,1))
maxImg = np.max(dc.data,axis=(2,3))
```

This should be safe so I'm PR'ing into the dev branch rather than the experimental IO branch.


bsavitzky:
Cool, thanks Steven!

PR #70, opened 2019-08-30
various changes to file handling

sezelt:
* Substantial changes to K2 binary handling: Now uses `np.memmap` instead of directly seeking around inside files, some further performance improvements. Fixes saving to v0.6 h5. 
* Substantial changes to h5 file handling. Files are now memory mapped, rather than preloaded into RAM. This causes some weirdness, which is partially addressed. Adds functions for closing all references to a given h5 file, which helps avoid the biggest pitfall, which is that the h5 file is kept open as long as the datacube exists.
* I have not yet made h5 memory mapping optional. I will, though...
* Changes how ROI is handled in `get_average_probe_from_ROI` because the h5 memmap doesn't support advanced slicing (nor does the K2 array for that matter). There may be similar problems elsewhere, which I will eventually seek out and repair. 
* One snag people have encountered is that you should **not** use something like `np.zeros_like(dc.data[:,:,0,0])` on a K2 dataset because this requires loading the entire dataset. This cannot/will not be fixed. Use `np.zeros((dc.R_Nx,dc.R_Ny))` instead.
* Adds support for the K2 binary format to parallel Bragg disk detection. Parallel Bragg disk detection should give a good speedup even on a desktop because it allows for multiple simultaneous reads.
* Gatan K2 files are now compatible as sources for Dask arrays. Wrapping a K2 dataset in a Dask array allows efficient support for all Dask array functions (most numpy array functions are implemented).

### DIFF
--- a/py4DSTEM/file/io/append.py
+++ b/py4DSTEM/file/io/append.py
@@ -41,7 +41,16 @@ def append_from_dataobject_list(dataobject_list, filepath):
     browser.close()
 
     #### Open file for read/write ####
-    f = h5py.File(filepath,"r+")
+    try:
+        f = h5py.File(filepath,"r+")
+    except OSError as e:
+        print(e)
+        print('The file appears to be open elsewhere...')
+        print('This can occur if your datacube was read from a py4DSTEM h5 file.')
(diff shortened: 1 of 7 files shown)


bsavitzky:
Sorry for the very slow turnaround on the PR Steve - looks great.  I would like to make sure the memmapping for .h5 is optional at some point.  Thanks!

PR #74, opened 2019-09-09
Quick bugfix

bsavitzky:


sezelt:
What was the error here? This change turns formatted strings into literals...

bsavitzky:
The string formatting was breaking the install script on at least one
windows machine.  Not sure why exactly other than windows being windows-y,
handling strings poorly...it was throwing a syntax error here, so to get it
running for a tutorial yesterday I just eliminated the formatting f...

sezelt:
f strings were added in python 3.6, maybe they had an older version?

PR #77, opened 2019-10-03
Better exception handling with file reading

HamishGBrown:
Creates an error message and resets the viewer if a file cannot be opened rather than crashing.


HamishGBrown:
I'll try update my branch and then reattempt this pull request

PR #84, opened 2019-11-25
EMD v0,7 with support for Counted Data

sezelt:
# EMD v0,7 with support for Counted Data

This update primarily adds support for electron counted datasets in sparse representation with the creation of a new class of `DataObject`, the `CountedDataCube`, which provides a wrapper around the mechanics for converting lists of electron events into diffraction patterns on the fly. The `CountedDataCube` is now also a group type in the EMD v0.7 format. 

The `CountedDataCube` has the same general behavior as a `DataCube` and returns reconstructed diffraction patterns on calls to `dc.data[Rx,Ry,:,:]`. Basic slicing along the Q axes is supported, and no slicing along the R axes is supported directly. Creating the `CountedDataCube` with the `use_dask` flag set will wrap the data in a Dask array that supports arbitrary slicing and more natural numpy support, but I have found this has very large overhead, and should be avoided unless absolutely necessary. (In certain circumstances Dask is probably faster, I haven't dug into it...) The electron events are stored in a `PointListArray`<sup>*</sup> at `cdc.electrons` and are converted to DPs on-the-fly as called using `numba` with approx. 10&mu;s overhead. Currently binning is only handled by densifying the data, creating a regular `DataCube`. The `CountedDataCube` can also be created in memory-mapped mode by passing a properly structured HDF5 dataset as the `electrons` argument. 

Accompanying this is a change to the way `PointListArrays` are structured inside the EMD file, making them much faster to read/write and more space-efficient. A fairly large `PointListArray` that I used as test data went from 100 MB down to 15 MB, and reading went from 30 s down to 5 s. The EMD v0.7 `CountedDataCube` is similar to Kitware's h5 electron counting files: we store ragged variable-length numpy arrays inside a 2D h5 dataset. The primary difference is that EMD will (prefer to) use structured numpy arrays, so that any `PointListArray` can be used, enabling counted datasets with 2D indexing, or extra fields for... things? (Data imported from Kitware is kept in 1D format. I have included an importer for the current Kitware file format but the long-term plan will be for Kitware to export in EMD v0.7. I will update the EMD website with the new spec shortly.) In order to maintain compatibility with HDF5 outside of Python, the arrays can also be simple datatypes. A string array dataset in the EMD group indicates which of the named fields of the `PointListArray` are used, and unstructured data is assumed to be 1D indexed. 

### `tqdmnd`
A side change is the addition of a better progress bar, a modification of [tqdm](https://tqdm.github.io) that better supports multidimensional iteration, which I have called `tqdmnd`. Usage is:
```python
from py4DSTEM.process.utils import tqdmnd

dc = py4DSTEM.file.datastructure.DataCube(...)
for (Rx, Ry) in tqdmnd(dc.R_Nx,dc.R_Ny, desc='Doing Things', unit='DP'):
    DP = dc.data[Rx,Ry,:,:]
    <DO STUFF>
```
will output a progress bar like:

`Doing Things: 100%|██████████| 263168/263168 [01:18<00:00, 3354.58DP/s]`

that keeps track of the progress, the iterations per second, and the elapsed/remaining times. 

### Notes on Memory Mapping

`FileBrowser.get_dataobject(...)` now has a keyword argument `memory_map`, which if set to `True` causes `DataCube` and `CountedDataCube` objects **in EMD v0.7 files only** to be returned as memory maps. Handling of previous EMD versions is left unchanged<sup>&dagger;</sup>. Memory mapping of `CountedDataCube` objects bypasses the creation of a PointListArray. On machines without a lot of available RAM, allocating the `PointListArray` can be very slow (minutes!), so memory mapping is preferred. Access performance for the `CountedDataCube` memory maps is much slower for each small read (250&mu;s per DP versus 10&mu;s in RAM) but still far faster than most operations on the data.

**When memory mapping data from the FileBrowser, you must keep the FileBrowser object open (i.e. you can't run `fb.close()` without losing access to the memory mapped data).** This creates roadblocks when trying to append to open data files, which was previously addressed in the `io_handling` branch. I have merged the improvements to K2 data reading and the GUI cursor and will abandon the rest of that branch. (Thus this PR supercedes the open PR #79).

When saving `CountedDataCubes`, saving from a memory map is not supported. The expectation here (which may change!) is that any manipulations worth saving to a new file (binning, densifying, etc.) will produce a new `DataObject`, that supports writing to EMD. 

### Compressed `DataCubes`
A new flag is added to the save function to enable native HDF5 compression on `DataCubes`: `py4DSTEM.file.io.save(...,compression=True)`. This enables chunking (1 DP per chunk) and `gzip` compression. This makes the files way smaller (I've seen it make 30x smaller files) but makes I/O way slower.

### GUI changes:

* Changes how the windows are sized. This has been a problem on Windows because the sizes we specified were too big, and you couldn't reach the whole window (it was off the screen). It became a much bigger problem with macOS Catalina, where it would segfault on launch if the window size was set in the code. This is probably ultimately a Qt bug...

#### Notes:

<sup>*</sup> When the `CountedDataCube` operates in memory-mapped mode, `cdc.electrons` is simply the `h5py.Dataset` from the EMD v0.7 file. For datasets read with the Kitware memory-mapped reader, the underlying dataset is 1D and cannot be reshaped on disk, so the index computation is done on the fly. Because there are so many different ways to create the `Sparse4D` object its state is rather complex, so it's dangerous to try to alter it once created. 

<sup>&dagger;</sup> This is a departure from the `io_handling` branch, where I had *all* datasets memory-mapped. This should be "safer."


py4DSTEM/file/datastructure/datacube.py
```diff
@@ -211,5 +230,192 @@ def get_virtual_image_annular_CoMY(self,slice_x,slice_y,R):
...
+            tf = TemporaryFile()
+            data4D = np.memmap(tf,dtype,'r+',shape=(newRx,newRy,newQx,newQy))
+        else:
+            data4D = np.zeros((newRx,newRy,newQx,newQy),dtype=dtype)
+
+        for (Rx, Ry) in tqdmnd(self.R_Nx, self.R_Ny, desc="Creating dense DC"):
+            rx = Rx // bin_R
+            ry = Ry // bin_R
+
+            data4D[rx,ry,:,:] += bin2D(self.data[Rx,Ry,:,:],bin_Q,dtype=dtype)
```
bsavitzky: Consider a conditional statement here that skips running bin2D() if bin_q==1

sezelt: 👍

py4DSTEM/file/io/append.py
```diff
@@ -34,18 +34,29 @@ def append_from_dataobject_list(dataobject_list, filepath):
...
 
     #### Open file for read/write ####
-    f = h5py.File(filepath,"r+")
+    try:
+        f = h5py.File(filepath,"r+")
+    except OSError as e:
+        print(e)
+        print('The file appears to be open elsewhere...')
+        print('This can occur if your datacube was read from a py4DSTEM h5 file.')
+        print(f'To forse close the file, losing any dataobjects open from it, run: py4DSTEM.file.io.close_h5_at_path(\'{filepath}\')')
```
bsavitzky: "Force"

bsavitzky:
MERGED AT LAST

PR #95, opened 2020-01-23
fix for reading EMD files

sezelt:
some untoward merges on my part made the last PR do things it should not do. this makes it do the things i said it would do.


bsavitzky:
How untoward!

PR #97, opened 2020-01-30
Amorphous Strain Mapping

tcpekin:


tcpekin:
This PR should add some functionality to nominally map strain from amorphous specimens. It uses the double-gaussian based fitting functions previously developed, and applies them to a stack. It then uses the coefficients, and computes the strain, as defined in Christoph's paper. 

There are a few things that should/will be updated in the future:
1. The implementation of Colin's latest strain derivation, once it is finalized
2. A switch to a different method of fitting the double sided gaussians, such that they are not overfit, and the use of r_ratio is not necessary. 
3. A proper implementation of A, B and C references (for if your reference 0 strain ellipse is not a perfect circle). The bones of this exists, but has not been tested. 
4. Potentially, despite my dislike, a working jupyter notebook instead of the strain_mapping.py script, that currently is used to process the data. 
5. The ability to make an elliptical correction to images, pre-fit. This would allow for the removal of any underlying strain/projector lens distortions before fitting a per pattern ellipse. This would most likely be done using the coefficients from the mean DP, and then used as a correction on a per-pattern basis. This would either be done using a function that generates a new corrected stack, or done immediately before the ellipse fit on a per-pattern basis. Perhaps then while this is happening, this stack is also saved. However, the implementation of this function I do not believe currently exists, with correct interpolation of full frames. I think it only exists for point list arrays.

Longer term, I would like to add symmetry mapping in the diffraction patterns and a function that computes a Gaussian weighted nearest neighbor stack (sum NN with Gaussian weights, to increase SNR for better fits)

bsavitzky:
Thanks Tom!  This looks great.  I'm going to accept the PR now.  A few comments:

- re: generating ellipse corrected diffraction patterns - this could def be worth doing at some point, but it would be somewhat computationally expensive, and it seems to me that for everything we've done so far, correctly accounting for a reference is good enough and much faster.  I probably won't try to do this until there is some application for which its necessary.
- symmetry mapping: I've just built some code to do symmetry analysis yesterday!  It's pretttttty neat.  Not mapping yet, but symmetry metrics are being calculated, so mapping is a hop-skip-n'-jump away.  Message me if you'd like me to send it along as is now; some time in the near-ish future it'll get rolled into a general update.
- Gaussian weighted NN DPs - this is a good idea which would be great to implement!  It would probably go elsewhere in the repo - I'm thinking in process/preprocess.
- Generally I might make a few changes to where some of these pieces of code live.  E.g. amorphous strain mapping can be its own module, rather than living in the rdf module.  Visualization code, like compare_double_sided_gaussian, I've been trying to keep out of files containing core analysis functions, so I may move this, though not 100% sure where yet, and this is an awfully handy visualization tool.  Your make_mask_array, which you noted could maybe live elsewhere, I'll probably put in utils.

Let me know if any of that doesn't sound right to you and I can hold off those edits, otherwise I'll start shuffling stuff around a bit.

Ben

tcpekin:
All of the proposals make sense to me - 
wrt elliptical correction - agreed, computationally expensive, and would much rather have a proper form for accounting for an elliptical reference. Exactly what the equation should be I'm not entirely sure about yet, though. Something like strain = 1/2(A-A_ref) or something, but I haven't verified this.

Symmetry mapping I'm definitely interested in - this is where elliptical correction makes the most sense, but if you do it via the polar transform, you can do that while converting to the polar transform, which I have seen that you already do. But would love to see that implemented. 

Where the code goes is completely up to you, and I'm fine with them moving anywhere. Conceptually, you have a much better idea of what is going on and what goes where than I do, so as long as I can find it, I'm happy.

Tom

PR #98, opened 2020-02-03
Pass filter functions into Bragg Disk detection routines and auto visualization

sezelt:
This adds the ability to pass a filtering function to the Bragg disk detection routines, which is lazily applied to each diffraction pattern. This is mainly motivated by the desire to bin data as part of the distributed bragg disk detection routine, where you want each worker to perform the filtering. The `filter_function` argument takes any callable of a single argument and applies it to each diffraction pattern before the correlation step. This can be a binning function, a filter such as Sobel, or something more complex like the exit wave power cepstrum. The default is `None`, which leaves the DPs untouched. 

I have also added a canned routine for visualizing the results of `find_Bragg_disks_selected` that takes any number of selected positions. 

Since I don't have the data for all the different workflow notebooks I'll just put a usage example here, where the DPs are binned by 4 before detecting the Bragg disks.

```python
# make a lambda that bins the DP by 4
binf = lambda x: py4DSTEM.process.utils.bin2D(x,4)

# the probe kernel must also be binned (if the sizes don't match an exception gets raised)
probe_kernel_binned = py4DSTEM.process.braggdiskdetection.get_probe_kernel(binf(probe))

# make a virtual image to show the realspace positions on
# (if none is given, a blank field with the right size is shown)
BF_image = np.max(dc.data,axis=(2,3))

######### find the peaks and display Bragg disks ########
# Select scan positions
Rxs=(240,120,133,21,300,30,180,10)
Rys=(90,90,302,400,299,503,500,10)

# Peak detection parameters
corrPower = 0.8
sigma = 2
edgeBoundary = 20
maxNumPeaks = 70
minPeakSpacing = 50
minRelativeIntensity = 0.001
subpixel = 'poly'

# Find peaks
peaks = py4DSTEM.process.braggdiskdetection.find_Bragg_disks_selected(dc, probe_kernel_binned, Rxs, Rys,
                                  corrPower=corrPower,
                                  sigma=sigma,
                                  edgeBoundary=edgeBoundary,
                                  minRelativeIntensity=minRelativeIntensity,
                                  minPeakSpacing=minPeakSpacing,
                                  maxNumPeaks=maxNumPeaks,
                                  subpixel=subpixel,
                                  filter_function = binf,
                                  show_image = BF_image)
```
Result:
im


bsavitzky:
Hey Steven-

This looks great, thanks!  The pre-filtering should be very useful to compare methods.

One issue: I've generally tried to keep plotting / visualization out of the processing files, and would prefer not to have the display kwargs/code inside the bragg disk detection routines.  This is a bit of a slippery slope - we could include this sort of visualization with just about every function, and the result of stuffing this stuff directly into the functional processing code could get sprawling and messy pretty fast.  If you feel strongly about having this visualization code included in the base code, the way I'd like to do this would be defining a subset of canned visualization routines inside their own submodule, and which can optionally be invoked in function calls, so that we only adding a small if statement in each processing function, rather than a big block of visualization code.  Whatchoo think?

sezelt:
I agree that having the viz code inside the analysis files can get unruly. 
So I've moved this new plot function and the generic `plot(...)` that Philipp added into a separate file in the `process.utils` submodule. 

I think that in this case it makes sense to keep the plotting as part of the selected Bragg disks function call, because that function exists mainly as a convenience wrapper for testing and visualizing parameters.

PR #139, opened 2021-03-19
small GUI and K3 DM fixes

sezelt:
This PR adds the ability to pass a filename as a command line argument when launching the GUI to immediately load that file, i.e. 
```
py4DSTEM /path/to/some/lovely/data/dm4
``` 
will open that lovely data in the GUI.

This also adds a fix for consolidated K3-produced `dm4` files. Such file contain the preview image, auto-generated VBF image, and a diffraction pattern, along with the 4D data in one `dm4` file. The K3 has been configured to not consolidate files by default anymore, but there are some such files still floating around in the wild. This fix is very old, and has been floating around as a patch with a few people, but I am finally submitting it for merger.

### DIFF
--- a/py4DSTEM/io/nonnative/read_dm.py
+++ b/py4DSTEM/io/nonnative/read_dm.py
@@ -6,6 +6,7 @@ from ncempy.io import dm
 from ..datastructure import DataCube, Metadata
 from ...process.utils import bin2D
 
+
 def read_dm(fp, mem="RAM", binfactor=1, metadata=False, **kwargs):
     """
     Read a digital micrograph 4D-STEM file.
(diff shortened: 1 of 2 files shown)


py4DSTEM/io/nonnative/read_dm.py
```diff
@@ -28,39 +29,70 @@ def read_dm(fp, mem="RAM", binfactor=1, metadata=False, **kwargs):
...
-    assert(binfactor>=1), "Error: binfactor must be >= 1"
+    assert isinstance(
+        fp, (str, Path)
+    ), "Error: filepath fp must be a string or pathlib.Path"
+    assert mem in [
+        "RAM",
+        "MEMMAP",
+    ], 'Error: argument mem must be either "RAM" or "MEMMAP"'
+    assert isinstance(binfactor, int), "Error: argument binfactor must be an integer"
+    assert binfactor >= 1, "Error: binfactor must be >= 1"
```
bsavitzky, line 40: Why this change?

py4DSTEM/io/nonnative/read_dm.py
```diff
@@ -28,39 +29,70 @@ def read_dm(fp, mem="RAM", binfactor=1, metadata=False, **kwargs):
...
+            # loop through the datasets until a >2D one is found:
+            i = 0
+            valid_data = False
+            while not valid_data:
+                data = dmFile.getMemmap(i)
+                if len(np.squeeze(data).shape) > 2 :
+                    valid_data = True
+                    dataSet = dmFile.getDataset(i)
+                i += 1
+            dc = DataCube(data=dataSet["data"])
```
bsavitzky, line 57: This is a meaningful change.  Will this not break if we have a 3D dataset?  If we're going to add this step, why not check to see if the data is 4D?  Is this to support 4DSTEM .dm files which have unraveled (Rx,Ry) into one long dimension?  They changed that a while back, yes?  I guess I'm wondering if we need to keep supporting this..... :/  I suppose we probably do....what do you think?

py4DSTEM/io/nonnative/read_dm.py
```diff
@@ -75,19 +107,16 @@ def get_metadata_from_dmFile(fp):
...
         pixelSizes = dmFile.scale
         pixelUnits = dmFile.scaleUnit
-        assert(pixelSizes[0]==pixelSizes[1]), "Rx and Ry pixel sizes don't match"
-        assert(pixelSizes[2]==pixelSizes[3]), "Qx and Qy pixel sizes don't match"
-        assert(pixelUnits[0]==pixelUnits[1]), "Rx and Ry pixel units don't match"
-        assert(pixelUnits[2]==pixelUnits[3]), "Qx and Qy pixel units don't match"
+        assert pixelSizes[0] == pixelSizes[1], "Rx and Ry pixel sizes don't match"
+        assert pixelSizes[2] == pixelSizes[3], "Qx and Qy pixel sizes don't match"
+        assert pixelUnits[0] == pixelUnits[1], "Rx and Ry pixel units don't match"
+        assert pixelUnits[2] == pixelUnits[3], "Qx and Qy pixel units don't match"
```
bsavitzky, line 113: Changed just for funsies?

bsavitzky (commented):
Thanks Steve.  Overall looks good!  Lmk what you think about my comments.

py4DSTEM/io/nonnative/read_dm.py
sezelt: This is a pure formatting change from running [black](https://black.readthedocs.io/en/stable/the_black_code_style.html) on the file

py4DSTEM/io/nonnative/read_dm.py
sezelt: This is also an auto-format from running black

py4DSTEM/io/nonnative/read_dm.py
sezelt: Aha yes this is a weird looking block, but the reason for it is that in the K3's consolidated `dm4` files, it saves a preview diffraction pattern with size `(1,1,Q_Nx,Q_Ny)`, hence the `squeeze`. 

In the Titan's `dm4` files, the data is saved as a 3D sequence, which (I have double-checked) is correctly chosen by this. These 3D datasets are still routinely produced.

py4DSTEM/io/nonnative/read_dm.py
bsavitzky: Hm, mkay.  Can you change this back please?  This looks silly and less readable to me.

py4DSTEM/io/nonnative/read_dm.py
bsavitzky: Gotcha.  Sigh!  Ok, looks good to me then

py4DSTEM/io/nonnative/read_dm.py
sezelt: This is rather odd looking, I admit, so I've reverted it

PR #140, opened 2021-03-19
filter functions for Bragg disk detection

sezelt:
This PR resurrects a focused subset of the changes originally proposed in PR #98, which was previously abandoned.

It adds an optional `filter_function` argument to the Bragg disk detection functions which operates on each DP immediately between slicing/loading the DP from the datacube and performing the cross-correlation. A salient example is for binning diffraction patterns on-the-fly before disk detection (which requires that the user supply a probe kernel of the appropriate size):

```python
# make a function that bins the DP by 4
binf = lambda x: py4DSTEM.process.utils.bin2D(x,4)

# the probe kernel must also be binned (if the sizes don't match an exception gets raised)
probe_kernel_binned = py4DSTEM.process.braggdiskdetection.get_probe_kernel(binf(probe))

######### find peaks ########

# Peak detection parameters
corrPower = 0.8
sigma = 2
edgeBoundary = 20
maxNumPeaks = 70
minPeakSpacing = 50
minRelativeIntensity = 0.001
subpixel = 'poly'

# Find peaks
peaks = py4DSTEM.process.diskdetection.find_Bragg_disks(dc, probe_kernel_binned,
                                  corrPower=corrPower,
                                  sigma=sigma,
                                  edgeBoundary=edgeBoundary,
                                  minRelativeIntensity=minRelativeIntensity,
                                  minPeakSpacing=minPeakSpacing,
                                  maxNumPeaks=maxNumPeaks,
                                  subpixel=subpixel,
                                  filter_function = binf)
```

In addition, the exit wave power cepstrum method can be applied using the newly added EWPC function in `utils`. Other potential uses are Sobel filtering, intensity normalization, background subtraction, etc.

### DIFF
--- a/py4DSTEM/process/diskdetection/diskdetection.py
+++ b/py4DSTEM/process/diskdetection/diskdetection.py
@@ -281,7 +281,6 @@ def find_Bragg_disks_selected(datacube, probe, Rx, Ry,
     t0 = time()
     for i in range(len(Rx)):
         DP = datacube.data[Rx[i],Ry[i],:,:] 
-
         peaks.append(_find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
                                                    corrPower = corrPower,
                                                    sigma = sigma,
--- a/py4DSTEM/process/diskdetection/diskdetection.py
+++ b/py4DSTEM/process/diskdetection/diskdetection.py
@@ -14,7 +14,7 @@ from ...io.datastructure import PointList, PointListArray
 from ..utils import get_cross_correlation_fk, get_maxima_2D, print_progress_bar, upsampled_correlation
 from ..utils import tqdmnd
 
-def find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
+def _find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
                                   corrPower = 1,
                                   sigma = 2,
                                   edgeBoundary = 20,
@@ -24,6 +24,7 @@ def find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
                                   maxNumPeaks = 70,
                                   subpixel = 'multicorr',
                                   upsample_factor = 16,
+                                  filter_function = None,
                                   return_cc = False,
                                   peaks = None):
     """
@@ -70,6 +71,12 @@ def find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
                                             'multicorr': uses the multicorr algorithm with
                                                         DFT upsampling
         upsample_factor      (int) upsampling factor for subpixel fitting (only used when subpixel='multicorr')
+        filter_function      (callable: lambda or partial function of one argument) filtering function to apply
+                             to each diffraction pattern before peakfinding. must be a function of only
+                             one argument (the diffraction pattern) and able to be pickled by the dill library.
+                             This is useful for doing noise reduction or binning on the fly for memory mapped
+                             datasets or for parallel operation. The shape of the returned DP must match the shape
+                             of the probe kernel.
         return_cc            (bool) if True, return the cross correlation
         peaks                (PointList) For internal use.
                              If peaks is None, the PointList of peak positions is created here.
@@ -80,6 +87,8 @@ def find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
         peaks                (PointList) the Bragg peak positions and correlation intensities
     """
     assert subpixel in [ 'none', 'poly', 'multicorr' ], "Unrecognized subpixel option {}, subpixel must be 'none', 'poly', or 'multicorr'".format(subpixel)
+
+    DP = DP if filter_function is None else filter_function(DP)
 
     if subpixel == 'none':
         cc = get_cross_correlation_fk(DP, probe_kernel_FT, corrPower)
@@ -153,10 +162,11 @@ def find_Bragg_disks_single_DP(DP, probe_kernel,
                                maxNumPeaks = 70,
                                subpixel = 'multicorr',
                                upsample_factor = 16,
+                               filter_function = None,
                                return_cc = False):
     """
-    Identical to find_Bragg_disks_single_DP_FK, accept that this function accepts a probe_kernel in
-    real space, rather than Fourier space. For more info, see the find_Bragg_disks_single_DP_FK
+    Identical to _find_Bragg_disks_single_DP_FK, accept that this function accepts a probe_kernel in
+    real space, rather than Fourier space. For more info, see the _find_Bragg_disks_single_DP_FK
     documentation.
 
     Accepts:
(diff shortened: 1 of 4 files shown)


py4DSTEM/process/diskdetection/diskdetection.py
```diff
@@ -70,6 +71,12 @@ def find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
                                             'multicorr': uses the multicorr algorithm with
                                                         DFT upsampling
         upsample_factor      (int) upsampling factor for subpixel fitting (only used when subpixel='multicorr')
+        filter_function      (callable: lambda or partial function of one argument) filtering function to apply
+                             to each diffraction pattern before peakfinding. must be a function of only
+                             one argument (the diffraction pattern) and able to be pickled by the dill library.
+                             This is useful for doing noise reduction or binning on the fly for memory mapped
+                             datasets or for parallel operation. The shape of the returned DP must match the shape
+                             of the probe kernel.
```
bsavitzky, line 79: A few things!
- Is there a reason we're not just calling this a function?  I'm not sure what is meant by "partial function of one argument" here - do you mean that it only accepts one argument?  If so this is specified later in the docstring.  It also seems to me like it shouldn't really matter if it's a lambda function or not.  I think the language here might be unnecessarily confusing. 
- pickling is only relevant for distribution to parallel processing, yes?  Assuming this is true,this should also be specified, at the end of the docstring.  Same design principle as above: best to keep things simple wherever possible, start with simple cases, and advance to more complex ones.  I suspect this would trip up plenty of users who are working locally and don't know what this means.
- "This is useful for..." I would leave this exposition out of the docstring.  We can show this in examples.

E.g. I'd suggest:
```
(function) filtering function to apply
 to each diffraction pattern before peakfinding. must be a function of only
 one argument (the diffraction pattern), which returns a diffraction pattern.
The shape of the returned DP must match the shape of the probe kernel
(important e.g. if the filter function is used for binning).  If used with parallelized
disk detection, the function must be able to be pickled by the dill library.
```

py4DSTEM/process/diskdetection/diskdetection.py
```diff
@@ -180,14 +190,21 @@ def find_Bragg_disks_single_DP(DP, probe_kernel,
...
```
bsavitzky, line 198: See comment above

py4DSTEM/process/diskdetection/diskdetection.py
```diff
@@ -237,12 +256,19 @@ def find_Bragg_disks_selected(datacube, probe, Rx, Ry,
...
```
bsavitzky, line 264: samesies

py4DSTEM/process/diskdetection/diskdetection.py
```diff
@@ -251,8 +277,9 @@ def find_Bragg_disks_selected(datacube, probe, Rx, Ry,
     # Loop over selected diffraction patterns
     t0 = time()
     for i in range(len(Rx)):
-        DP = datacube.data[Rx[i],Ry[i],:,:]
-        peaks.append(find_Bragg_disks_single_DP_FK(DP, probe_kernel_FT,
+        DP = datacube.data[Rx[i],Ry[i],:,:] 
+
```
bsavitzky, line 281: why!?

py4DSTEM/process/diskdetection/diskdetection.py
```diff
@@ -319,6 +348,12 @@ def find_Bragg_disks_serial(datacube, probe,
                                 to the median of the maximum intensity peaks in each diffraction pattern. 'manual' Allows the user to threshold
                                 based on a predetermined intensity value manually determined. In this case, minIntensity should be an int.
         verbose              (bool) if True, prints completion updates
+        filter_function      (callable: lambda or partial function of one argument) filtering function to apply
+                             to each diffraction pattern before peakfinding. must be a function of only
+                             one argument (the diffraction pattern) and able to be pickled by the dill library.
+                             This is useful for doing noise reduction or binning on the fly for memory mapped
+                             datasets or for parallel operation. The shape of the returned DP must match the shape
+                             of the probe kernel.
```
bsavitzky, line 356: same same same

py4DSTEM/process/diskdetection/diskdetection.py
```diff
@@ -404,6 +446,13 @@ def find_Bragg_disks(datacube, probe,
                                                         DFT upsampling
         upsample_factor      (int) upsampling factor for subpixel fitting (only used when subpixel='multicorr')
         verbose              (bool) if True, prints completion updates for serial execution
+        filter_function      (callable: lambda or function of one argument) filtering function to apply
+                             to each diffraction pattern before peakfinding. must be a function of only
+                             one argument (the diffraction pattern) and able to be pickled by the dill library.
+                             This is useful for doing noise reduction or binning on the fly for memory mapped
+                             datasets or for parallel operation. The shape of the returned DP must match the shape
+                             of the probe kernel. When using distributed mode, the function must contain
+                             all its own imports.
```
bsavitzky, line 455: the fact that it was necessary to add this 17 times is evidence of poor code architecture.  My bad :/

py4DSTEM/process/diskdetection/diskdetection_parallel.py
```diff
@@ -1,11 +1,11 @@
 # stdlib
 import os
-import pickle
 import tempfile
 from time import time
 
 # 3rd party
 import numpy as np
+import dill
```
bsavitzky, line 8: Just confirming - this should not be a new dependency, `dill` is included in base python past 3.whatever, right?

py4DSTEM/process/utils/utils.py
```diff
@@ -548,8 +548,14 @@ def get_voronoi_vertices(voronoi, nx, ny, dist=10):
 
     return vertex_list
 
-
-def foo():
-    """Test func."""
-    print('pass')
+def ewpc_lambda(Q_Nx,Q_Ny):
```
bsavitzky, line 551: I would rename this - similar to my comment above, I don't see how being a lambda function is relevant.  Isn't the relevant info is that this is returning a function?  This name seems to obfuscate more than it helps - I'd recommend calling this `get_ewpc_filter_function` if you want to be explicit, or `get_ewpc_fn` if you want to be concise.

bsavitzky (changes requested):
Thanks Steve - very useful update, and I'm glad to finally get it rolled in.

The main take home of most of my comments is that, as a general rule, I have a strong preference for keeping docstrings/descriptions as simple as possible, starting with simpler descriptions first, and then adding complexity later and only when its necessary and adds value.  Saying things in a technically precise way first may make everything clear as concisely as possible to an advanced user, but it also can make it difficult or unreadable to beginning users.  More specific suggestions are in my comments!

Apropos of a couple of my comments, I would probably call this update "filter functions for Bragg disk detection", as it doesn't seem to me like it matters if the function you pass is defined using lambda syntax, or some other way.  If I'm missing something here, please let me know.

py4DSTEM/process/diskdetection/diskdetection.py
sezelt: i'm.... not sure 🤷

sezelt:
I've updated the many many many places where the docstring appears, to hopefully be more appropriate. 

`dill` seems to be as good as part of the standard library, but I've added it to the `setup.py` to be sure that it's present.

PR #141, opened 2021-03-19
add back K2 binary reading

sezelt:
This PR re-adds the ability to read Gatan's proprietary K2 `gtg`/`bin` format, within the new `io` framework.

A small but important change from before is that the code no longer uses `os.chdir`, which used to cause bizarre errors if you opened a K2 dataset on a remote volume.

Users can pass any of (i) the path to the `gtg` file, (ii) the path to any of the `bin` files, or (iii) the path to the folder containing the `gtg` and `bin` files, but only (i) or (ii) results in py4DSTEM auto-detecting the appropriate reader. To use (iii), supply `ft="gatan_K2_bin"` to `read`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ dist/
 py4DSTEM/test/testdata/
 sample_code/data/*h5
 sample_code/data/*dm3
+.vscode
 # Data Specific #
 #################
 *.png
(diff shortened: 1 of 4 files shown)


bsavitzky:
Ok!  There are some formatting changes made by black that I can't say I love here, but....as discussed, I think perhaps in the interest of uniformity, beginning to use this is fine.  Everything else looks great, again, as per our conversation yesterday.

PR #147, opened 2021-05-05
Amorphous code updated, including strain.

tcpekin:
Here is my request, probably will have to update things, but there is a lot to go through here potentially, so thought I would send it and give you some time to look through it. To my knowledge.. it all works.


bsavitzky (commented):
wooooooooo

PR #148, opened 2021-05-13
Changes 'center' -> 'origin' in center_braggpeaks 

abruefach:
fix minor bug due to name change of " .get_center "


bsavitzky (approved):
sweeeeeeeet ty Alex!

PR #163, opened 2021-06-07
Vacuum probe template

cophus:
This PR contains a tutorial notebook for constructing probe templates. It covers 4 use cases, and links to example data in the google drive.

I also fixed the probe masking, but replacing the hard marching squares edges with a soft edge based on the binary Euclidean distance transform.  Should be better in all cases.


bsavitzky:
The main repo changes here have now been moved and merged in PR 164, and the notebook edits will happen in the new tutorials repo.

PR #165, opened 2021-06-14
Gdrive downloader

alex-rakowski:
added google drive download util and updated init file


py4DSTEM/io/google_drive_downloader.py
```diff
@@ -0,0 +1,44 @@
+import requests 
+from typing import Union
+import pathlib
+
+
+def download_file_from_google_drive(id_:str, 
+                                    destination:Union[pathlib.PurePosixPath, pathlib.PureWindowsPath,str]) -> None:
```
bsavitzky, line 7: This style of fn/arg/return definition isn't consistent with the rest of the code, and the same info is included in the docstring you're written below (and will be scraped by sphinx for the docpages), so just for consistency I'd recommend changing to
```
def download_file_from_google_drive(id_,destination]):
```

bsavitzky (commented):
Looks good to me!  By which I mean I don't understand the internet, so, if you tell me this is good enough, I will trust that it is ;p

bsavitzky (approved):

PR #166, opened 2021-06-14
Fixing some .show inconsistencies

cophus:
When selecting different image scaling modes, such as manual, std, log or power, we want to keep the intensity range / clipvals as self consistent as possible.

For example before this update:
image

And after this update:
image

Note how we can keep the intensity ranges the same and produce the same images for "manual" and "std."  Similar logic has been applied to log and power.  Finally, if the min=0 setting is used for log scaling, this value is instead reset to the minimum value of the image after taking the log.

### DIFF
--- a/py4DSTEM/visualize/show.py
+++ b/py4DSTEM/visualize/show.py
@@ -161,17 +161,21 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
         _mask = ar>0
         _ar = np.zeros_like(ar.data,dtype=float)
         _ar[_mask] = np.log(ar[_mask])
-        if min != 0:
-            min = np.log(min)
-        else:
-            min = np.min(_ar[_mask])
-        max = np.log(max)
+        if min != None:
+            if min != 0:
+                min = np.log(min)
+            else:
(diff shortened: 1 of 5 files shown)


py4DSTEM/visualize/show.py
```diff
@@ -161,10 +161,21 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
         _mask = ar>0
         _ar = np.zeros_like(ar.data,dtype=float)
         _ar[_mask] = np.log(ar[_mask])
+        if min != None:
+            if min != 0:
```
bsavitzky, line 165: What if min is negative? Perhaps we want ``if min > 0``?

bsavitzky (commented):
All these edits look good to me, with the one comment I made in the log scaling code where I think there may be a small bug.

I do have some uncertainty about this approach to the clipvals+scaling.  I understand it's benefit, and it is significant.  That said, I am concerned this will also make tuning clipvalues for scaled data rather difficult.  Not sure if there's a workaround that preserves the nice quality of consistency across scaling.

bsavitzky (approved):

PR #174, opened 2021-06-23
Add support for reading EMD using MEMMAP

yo-my-bard:
Hello! I noticed there was a TODO to add MEMMAP support for EMD datasets as well. As far as I could tell, it didn't seem like it was supported quite yet.

I made an attempt to do so by mirroring the logic that exists for read_dm.py.

## Testing
- This was tested using ~32 and ~70GB h5 files on a 48GB RAM computer and worked pretty well. 
- I tested it with [the provided small h5 file](https://drive.google.com/file/d/12Q3T57x9N2vkyY0llqBLKn_0JPurQM6Y/view?usp=sharing) and had success there. 
- I added a small unit test to note that there shouldn't be regression: DataCube result should be the same whether using MEMMAP or RAM.

I'm less familiar with the various technical use cases of the library and whether we'd need to generalize this approach further, but  I'm happy to continue to work on the PR and address any comments that come up!

Thank you for your time!

### DIFF
--- a/py4DSTEM/io/datastructure/datacube.py
+++ b/py4DSTEM/io/datastructure/datacube.py
@@ -10,7 +10,8 @@ import numpy as np
 import numba as nb
 import h5py
 
-import numpy as np
+from ncempy.io import emd
+
 from .dataobject import DataObject
 from ...process import preprocess
 from ...process import virtualimage_viewer as virtualimage
@@ -286,7 +287,15 @@ def get_datacube_from_grp(g,mem='RAM',binfactor=1,bindtype=None):
     """ Accepts an h5py Group corresponding to a single datacube in an open, correctly formatted H5 file,
         and returns a DataCube.
     """
-    # TODO: add memmapping, binning
+    # TODO: add binning
+    if mem == 'MEMMAP':
+        h5_filename = g.file.filename
+        with emd.fileEMD(h5_filename) as emdF:
+            emd_memmap = emdF.get_memmap(0)
+            dataset = emd_memmap[0]
+            mm = np.memmap(h5_filename, shape=dataset.shape, dtype=dataset.dtype, offset=dataset.id.get_offset())
+            name = dataset.parent.name.split('/')[-1]
+            return DataCube(data=mm, name=name)
     data = np.array(g['data'])
     name = g.name.split('/')[-1]
     return DataCube(data=data,name=name)
--- a/py4DSTEM/io/datastructure/datacube.py
+++ b/py4DSTEM/io/datastructure/datacube.py
@@ -290,12 +290,10 @@ def get_datacube_from_grp(g,mem='RAM',binfactor=1,bindtype=None):
     # TODO: add binning
     if mem == 'MEMMAP':
         h5_filename = g.file.filename
-        with emd.fileEMD(h5_filename) as emdF:
-            emd_memmap = emdF.get_memmap(0)
-            dataset = emd_memmap[0]
-            mm = np.memmap(h5_filename, shape=dataset.shape, dtype=dataset.dtype, offset=dataset.id.get_offset())
-            name = dataset.parent.name.split('/')[-1]
-            return DataCube(data=mm, name=name)
+        emdF = emd.fileEMD(h5_filename)
+        dataset, dims = emdF.get_memmap(0)
+        name = dataset.parent.name.split('/')[-1]
+        return DataCube(data=dataset, name=name, nothing=emdF)
     data = np.array(g['data'])
     name = g.name.split('/')[-1]
     return DataCube(data=data,name=name)
--- a/py4DSTEM/test/test_io/test_datastructure/test_datacube.py
+++ b/py4DSTEM/test/test_io/test_datastructure/test_datacube.py
@@ -1,0 +1,51 @@
+import unittest
+import os
+import h5py
+import numpy as np
+
+from tempfile import mkdtemp
+from shutil import rmtree
+
+from py4DSTEM.io.datastructure.datacube import get_datacube_from_grp
+
+class Test(unittest.TestCase):
+
+    def setUp(self):
+        self.tmpdir = mkdtemp()
+        self.h5_file = h5py.File(os.path.join(self.tmpdir, 'test.h5'), 'w')
+        (self.h5_file
+         .create_group('4DSTEM_experiment')
+         .create_group('data')
+         .create_group('datacubes')
+         .create_group('datacube_0')
+         .create_dataset('data', data=np.ones(shape=(2,2,3,3)), dtype='uint8')
+        )
+        self.datacube = self.h5_file['4DSTEM_experiment']['data']['datacubes']['datacube_0']
+        self.datacube.attrs.create('emd_group_type', 1)
+        dim1 = self.datacube.create_dataset('dim1', shape=(2,), dtype='uint8')
+        dim1.attrs.create('name', np.string_("R_x"))
+        dim1.attrs.create('units', np.string_("[pix]"))
+        dim2 = self.datacube.create_dataset('dim2', shape=(2,), dtype='uint8')
+        dim2.attrs.create('name', np.string_("R_y"))
+        dim2.attrs.create('units', np.string_("[pix]"))
+        dim3 = self.datacube.create_dataset('dim3', shape=(3,), dtype='uint8')
+        dim3.attrs.create('name', np.string_("Q_x"))
+        dim3.attrs.create('units', np.string_("[pix]"))
+        dim4 = self.datacube.create_dataset('dim4', shape=(3,), dtype='uint8')
+        dim4.attrs.create('name', np.string_("Q_y"))
+        dim4.attrs.create('units', np.string_("[pix]"))
+        
+
+    def testH5Memmap(self):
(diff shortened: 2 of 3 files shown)


py4DSTEM/io/datastructure/datacube.py
```diff
@@ -10,7 +10,8 @@
 import numba as nb
 import h5py
 
-import numpy as np
```
yo-my-bard, line 13: Duplicate import

py4DSTEM/io/datastructure/datacube.py
```diff
@@ -286,7 +287,15 @@ def get_datacube_from_grp(g,mem='RAM',binfactor=1,bindtype=None):
...
         and returns a DataCube.
     """
-    # TODO: add memmapping, binning
+    # TODO: add binning
+    if mem == 'MEMMAP':
+        h5_filename = g.file.filename
+        with emd.fileEMD(h5_filename) as emdF:
+            emd_memmap = emdF.get_memmap(0)
+            dataset = emd_memmap[0]
+            mm = np.memmap(h5_filename, shape=dataset.shape, dtype=dataset.dtype, offset=dataset.id.get_offset())
```
yo-my-bard, line 296: Brilliant approach here: https://groups.google.com/g/h5py/c/1Fjt19OPcYY/m/HoXXJARwAgAJ

ercius: The memmap returned by ncempy's function `get_memmap()` uses h5py for accessing the dataset directly from disk. You are providing instead a numpy memmap to the binary chunk on disk. This requires that the data is contiguous on disk (i.e. the h5py dataset does not use chunking or filtering in h5py) to work properly.

It might be better to just allow h5py to provide the memmap instead, because then the user does not have to worry about data contiguity. h5py takes care of that for you. So, you can just return `emd_memmap[0]`.

yo-my-bard: Thanks for the feedback! That makes sense!
It looks like making this change poses challenges. The current expectation, as I understand it, is:
- Passing around a DataCube instance
- DataCube.data is an ndarray type

Given that the entrypoint for this logic is `py4dstem.io.read()`, returning `h5py.Dataset` for MEMMAP cases would break backward compatibility for anyone who chooses to pass in MEMMAP. 

- Does it make sense to change the purpose of this pull request so it's clear that this change would only support contiguous datasets for now? I could add some assertions to gate usage accordingly and continue thinking about a more complete solution for another pull request.
- Are there other approaches I could take that I'm missing? For e.g., maybe adding the logic here isn't right, but somewhere else might work?
  - Something else I noticed is that when I pass around the get_memmap() result, the DataSet is closed after returning the result from the function. Looks like it got garbage collected and closed accordingly.

I committed my latest attempt for visibility. I'm still learning my way around, so I appreciate the feedback!

ercius: > Given that the entrypoint for this logic is py4dstem.io.read(), returning h5py.Dataset for MEMMAP cases would break backward compatibility for anyone who chooses to pass in MEMMAP.

Im not sure this is true. Most things that are done in numpy can be done on a hdf5_dataset object equally. Not every function is allowed (i.e. reshape). So, it still might not break anything but is only a partial "drop-in" replacement.

> Does it make sense to change the purpose of this pull request so it's clear that this change would only support contiguous datasets for now? I could add some assertions to gate usage accordingly and continue thinking about a more complete solution for another pull request.

In my opinion: yes. But its up to maintainers of py4DSTEM to weigh in on how they want to support HDF5 memmaps. Im just providing context from the `ncempy` reading side which is being used under the hood in py4DSTEM.

> Something else I noticed is that when I pass around the get_memmap() result, the DataSet is closed after returning the result from the function. Looks like it got garbage collected and closed accordingly.

This is because you are using the `with` context manager. This is suggested to avoid accidentally leaving a file open in other cases. Not in this instance, however. Here you are opening the file, returning the hdf5_dataset object, and finally closing the file. For any memmap (hdf5, numpy, etc.) you have to have the file open to keep accessing it on disc. Remove the `with` to solve that problem.

ercius:
@sezelt just added EMD memmap support using the h5py.dataset instead of creating a np memmap. It was just merged. You should check out that new feature.

bsavitzky:
Hi @yo-my-bard - thanks for this contribution!

As @ercius mentioned, I just accepted a newer PR that added EMD memmapping almost the same way you did (#182).  The duplicated effort is my fault, I'm sorry I did not get to this PR sooner.  However, I think this is still quite helpful!  Two points -

- The unit testing code added in this PR is quite important, and should definitely be merged
- PR #182 did not use ncempy, choosing instead to just use h5py directly, whereas this PR uses ncempy

I think the best way to proceed would for me to add a commit to sort out the merge conflicts with PR#182, and then accept this PR.  The only unresolved question for me is - in the core few lines in `datacube.py` where there is a conflict, do we want to use ncempy, or just use the h5py.Dataset directly?  Pros/cons: just using h5py is simpler and works well enough for now.  Using ncempy makes things very slightly more tortuous/opaque, but has the advantage of getting the dim vector metadata, even if we're not using it (yet).  I think this last point is important, so perhaps using ncempy as is done here is better, but am curious what the author relevant authors think...?

@yo-my-bard @ercius @sezelt

ercius:
I'm my opinion, getting the metadata is very important. Getting that right
is always harder but worth it in the long run. I'm happy to try and push
anything into ncempy that is needed to support this.

bsavitzky:
Ok, this has now been merged, using ncempy to get the h5py.Dataset memmap following @yo-my-bard 's PR, and using the control flow logic from @sezelt 's PR.  This means dim vectors are now being read using ncempy in `Datacube.get_datacube_from_grp` @ercius.  They're not used yet - this will need to get added soon.  Thanks all :)

sezelt:
Hey @bsavitzky it seems this PR only edits `DataCube.get_datacube_from_grp`, but that function is only called for v0.12 EMD files. All lower version EMDs have other versions of this function in their `read_v_XX.py` files, and so are using the code from PR #182 rather than this change. It is probably safe to nix those other copies of the reader function and have those all call the one in `DataCube`, but I haven't confirmed this for certain. 

Also, why does this PR remove the `assert` in the `DataCube` version of the reader that will alert if the user asks for binning, which we are not yet supporting? Having this ignore inputs and misbehave silently will be confusing.

bsavitzky:
Thanks @sezelt.  That's true that this is called by v0.12 (and planned for v0.12+) files.  I was going to leave the older version EMD readers as they are now, as I think they PR#182 method should be fine - indeed, functionally should be identical to this, as ultimately both just make a datacube where the data is an h5py.Dataset.  The only real difference here is that we now have access to the dim vectors, which will be useful in the future, but as they aren't being used now, I don't see a real reason to change the earlier version datacube readers at present.  If you wanna change them, certainly feel free!

The assert statement was kept during merge conflict handling :p

yo-my-bard:
Thanks for looking into this and for the discussion here -- it's great to see support for memmap!

One follow-up I had was regarding h5py closing the dataset after Datacube is returned from `DataCube.get_datacube_from_grp`. I removed usage of the `with` context manager and this issue remains, I think. The unit test that was merged is an example.

It could be that I'm still not using it correctly even in the unit test (or the code for that matter). If everything is working as you expect based on your tests, then great! Just wanted to make a note of that. (Also would appreciate if you can share your test steps). Thanks!

sezelt:
@yo-my-bard I had a similar concern about how ncempy returns the memory map, but when I tested it on my machine (even invoking the garbage collector explicitly with `gc.collect()`) I did not have the problem with the dataset getting closed. I think you are right that in ncempy, when the `fileEMD` object is deleted its `__del__` method closes the file (regardless of whether there is an open memmap or not). @ercius this is probably not a problem for people using `fileEMD` directly, but when it's used like this inside another function where it will get garbage collected, it seems memory maps can get broken. 

In #182 I create a new `h5py.File` object outside the context manager in order to return the memory map, and I think that as long as the resulting `h5py.Dataset` is alive (which is not a problem, since the `DataCube` object maintains a reference to it) then the file will not be automatically closed. 

@bsavitzky can you let us know if the unit tests passed when you ran them? This does seem to be a slightly random issue.

bsavitzky:
Good catch, oof, yes, this is a problem!  I am getting a closed h5 file.  I missed this because I tested with a datacube that, it turns out, was not v0.12, so it was getting read using h5py directly following PR#182.  (Which is exactly the danger of having different reading methods that you pointed out, @sezelt ...alas).  Reading a v0.12 datacube with `mem=MEMMAP` is giving me a closed h5 dataset.  Arg!

For clarity - I've been pretty ad hoc about testing thus far...it is going to be important to switch over to proper unit testing soon, but for now (before/until MM2021), most of my testing has just been writing and manually running small scripts. :'/  I did run the unittests from this PR - they did not pass, but I'm not certain if this is an issue with the source or test code.  In any event - the issue with keeping the `h5py.Dataset` open remains.

As for solutions...your workaround to maintain a reference to the memmap may be the easiest aproach @yo-my-bard.  Certainly this is a good approach if we want to just handle this within py4DSTEM.  Curious if @ercius has any thoughts from the ncempy side?

PR #177, opened 2021-06-29
Add reader for EMPAD raw files

sezelt:
This PR adds the ability to read files from the original EMPAD (support for EMPAD 2 is unknown, as I don't have any test files for that). The reader supports reading into memory, memory-mapping, and bin-on-read to RAM. 

Metadata is not supported as there is no meaningful metadata that I am aware of in the file format. (A separate XML file is used for that metadata, but I have not added support for reading from that file. Perhaps this would be handled through a `kwarg` like `metadata_file` if we chose to implement it, since in the examples I've seen the XML does not have the same filename as the data)

### DIFF
--- a/py4DSTEM/io/nonnative/empad.py
+++ b/py4DSTEM/io/nonnative/empad.py
@@ -1,5 +1,5 @@
 # Reads an EMPAD file
-# 
+#
 # Created on Tue Jan 15 13:06:03 2019
 # @author: percius
 # Edited on 20190409 by bsavitzky and rdhall
@@ -7,6 +7,7 @@
 
 import numpy as np
 from pathlib import Path
+
 
 def read_empad(filename):
     """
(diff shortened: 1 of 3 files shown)


py4DSTEM/io/nonnative/empad.py
```diff
@@ -0,0 +1,91 @@
...
+# @author: percius
+# Edited on 20190409 by bsavitzky and rdhall
+# Edited on 20210628 by sez
+
+import numpy as np
+from pathlib import Path
+from ..datastructure import DataCube
+from ...process.utils import bin2D, tqdmnd
+
+from pdb import set_trace
```
bsavitzky, line 13: We can remove this, no?

py4DSTEM/io/nonnative/empad.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Accepts:
+        filename    (str) path to the empad file
+
+    Returns:
+        data        (ndarray) the 4D datacube, including the metadata.
+                    raw data only can be accessed data[:,:,:128,:]
+    """
+
+    # fmt: off
```
bsavitzky, line 35: rm? This is leftover from using Black for formatting?

py4DSTEM/io/nonnative/empad.py
```diff
@@ -0,0 +1,91 @@
...
+    assert(isinstance(filename, (str, Path))), "Error: filepath fp must be a string or pathlib.Path"
+    assert(mem in ['RAM', 'MEMMAP']), 'Error: argument mem must be either "RAM" or "MEMMAP"'
+    assert(isinstance(binfactor, int)), "Error: argument binfactor must be an integer"
+    assert(binfactor >= 1), "Error: binfactor must be >= 1"
+    assert(metadata is False), "Error: EMPAD Reader does not support metadata."
+    # fmt: on
```
bsavitzky, line 41: rm?

py4DSTEM/io/nonnative/empad.py
```diff
@@ -0,0 +1,91 @@
...
+        imFirst = np.fromfile(fid, dtype=empadDTYPE, count=1)
+        fid.seek(-128 * 130 * 4, 2)
+        imLast = np.fromfile(fid, dtype=empadDTYPE, count=1)
+
+    # Get the scan shape
+    shape0 = imFirst["metadata"][0][128 + 12 : 128 + 16]
+    shape1 = imLast["metadata"][0][128 + 12 : 128 + 16]
+    # kShape = shape0[2:4]  # detector shape (should always be row x col)
+    rShape = 1 + shape1[0:2] - shape0[0:2]  # scan shape
+
```
bsavitzky, line 59: rm blank line

py4DSTEM/io/nonnative/empad.py
```diff
@@ -0,0 +1,91 @@
...
+    size. Then, the full dataset is loaded.
+
+    Note that the output of this function is a datacube which includes the 2 rows of metadata.
+
+    Accepts:
+        filename    (str) path to the empad file
+
+    Returns:
+        data        (ndarray) the 4D datacube, including the metadata.
+                    raw data only can be accessed data[:,:,:128,:]
```
bsavitzky, line 32: It says here that the fn returns the datacube *with* the 2 metadata lines, but the code (lines 64,68,72) do *not* appear to be including the metadata lines in the returned datacube.  I think this is probably the right way to do this, but we should then change the `Note` and `Returns` descriptions in this docstring

bsavitzky (changes requested):
Looks good.  The only meaningful edit requested is to please confirm that this function does *not* include the 2 metadata lines in the output datacube, and to edit the docstring accordingly

sezelt:
@bsavitzky I agree with all your comments, and addressed them in this new commit. I edited the docstring to clarify that we do indeed return only the 128x128 valid data region, and also corrected that we are returning the `DataCube`, not a `np.ndarray`. Those other little comment lines were indeed there to prevent black from reformatting those assert statements to look insane, but since we in general aren't (yet!) running a formatter on the whole codebase I just nixed them.

bsavitzky:
> @bsavitzky I agree with all your comments, and addressed them in this new commit. I edited the docstring to clarify that we do indeed return only the 128x128 valid data region, and also corrected that we are returning the `DataCube`, not a `np.ndarray`. Those other little comment lines were indeed there to prevent black from reformatting those assert statements to look insane, but since we in general aren't (yet!) running a formatter on the whole codebase I just nixed them.

Thanks @sezelt - one last thing - I take it you don't still need to be importing the python debugger, yes?  Cool to remove this line?

sezelt:
Yes, it's safe to remove `pdb`.

bsavitzky (approved):

Issue #156, opened 2021-05-29
py4DSTEM GUI issue


PhysXu:
I followed the instruction in Github for installing py4DSTEM:

          conda update conda
          conda create -n py4dstem python==3.7
          conda activate py4dstem
          conda install pip
          pip install py4dstem
          conda install pywin32

But when I try to run the GUI by

          conda activate py4dstem
          py4dstem

It tells me:
![image](https://user-images.githubusercontent.com/35793078/120061920-09f8f180-c092-11eb-8f9d-4647e79f9a20.png)

alex-rakowski:
The GUI is currently broken, and we are working on the fix. We'll post an update here https://github.com/py4dstem/py4DSTEM#running-the-gui , when it is fixed.

Issue #133, opened 2020-12-03
I can't find the "Browse" button in the main window, and when I change the size of whole window, I still can't get this button.


yabowang-deeper:

alex-rakowski:
Assume you are referring to the GUI? If so there are known issues with it, we are working on.  We will post and update here https://github.com/py4dstem/py4DSTEM#running-the-gui when the GUI is functional again

bsavitzky:
@sezelt is this fixed in the current GUI?

sezelt:
yes

PR #179, opened 2021-06-29
Fixes for GUI

sezelt:
This PR cleans up some of the incompatibilities and old names in the GUI code. Should solve #156, and other issues reported in the Slack. 

Someone should actually test this before merging... On my laptop, every random datacube I have is in an old version of EMD which the reader now refuses to open, and on my desktop I somehow borked PyQt5. Leaving this as a draft for now, since it's not confirmed working.


sezelt:
After a bit more tweaking, the GUI now works again. There may be some residual bugs, in particular in the last step of strain mapping, but it appears OK in the testing I have done. 

I also made the "Browse" button left-aligned instead of right in hopes of making it more likely to show up on Windows computers with small screens, which has always been a problem (so this should hopefully solve #133 )

bsavitzky (approved):
Great, thanks Steve

PR #180, opened 2021-07-01
add dtype selection and progess bar to bin_data_mmap

sezelt:
A tiny tweak to allow choice of the target dtype when binning a memory mapped DataCube. Changes the for loop to use `tqdmnd` for progress updates.

### DIFF
--- a/py4DSTEM/io/datastructure/datacube.py
+++ b/py4DSTEM/io/datastructure/datacube.py
@@ -97,7 +97,7 @@ class DataCube(DataObject):
         self = preprocess.bin_data_diffraction(self, bin_factor)
 
     def bin_data_mmap(self, bin_factor, dtype=np.float32):
-        self = preprocess.bin_data_mmap(self, bin_factor)
+        self = preprocess.bin_data_mmap(self, bin_factor, dtype=dtype)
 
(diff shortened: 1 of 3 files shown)


sezelt:
I have added another tiny tweak, which is to use for loops in the virtual image functions, as a user encountered problems with running out of memory generating an annular image, even when using a memory-mapped dataset.

py4DSTEM/io/datastructure/datacube.py
```diff
@@ -96,7 +96,7 @@ def crop_data_real(self,crop_Rx_min,crop_Rx_max,crop_Ry_min,crop_Ry_max):
     def bin_data_diffraction(self, bin_factor):
         self = preprocess.bin_data_diffraction(self, bin_factor)
 
-    def bin_data_mmap(self, bin_factor):
+    def bin_data_mmap(self, bin_factor, dtype=np.float32):
         self = preprocess.bin_data_mmap(self, bin_factor)
```
bsavitzky, line 100: shouldn't line 100 have `dtype=dtype`?

bsavitzky (changes requested):

bsavitzky (approved):

PR #182, opened 2021-07-19
add MEMMAP support for EMD/HDF5 files

sezelt:
This adds memory mapping back to EMD files, by returning the `h5py.Dataset` object rather than reading the data to a numpy array, when the user specifies `mem="MEMMAP"` in `py4DSTEM.io.read()`.

I have confirmed that this works with the Ge/SiGe tutorial dataset and notebook without any problems, though it would be best to test this out with other versions of py4DSTEM file to be absolutely certain.


bsavitzky (approved):
Looks good, thanks Steve!

ercius:
This is probably the best way to go forward with respect to #174 also. Using the HDF5 dataset rather than a numpy memmap will provide more flexible memory mapping for all HDF5-type datasets.

PR #184, opened 2021-07-23
add absolute intensity to bragg disk detection

sezelt:
This PR adds an absolute correlation intensity threshold to the Bragg disk detection routines.

I have left the default thresholds to be `minRelativeIntensity=0.005`, `minAbsoluteIntensity=0`, so that the new feature will not change the behavior of any existing notebooks that fall back on the default; we may want to discuss what values we want to start out with, going forward.


bsavitzky (approved):
Looks good!

PR #185, opened 2021-07-29
more GUI fixes

sezelt:
This PR fixes bugs that affected reading EMD files in the GUI that remained after PR #179 . 

I have tested this on py4DSTEM EMD file versions 0.12, 0.9, 0.6, and 0.5, and it loads them correctly. I have deliberately only edited things in the GUI code to keep any potential new bugs contained, though a cleaner solution integrated into the rest of the reader functions is probably warranted in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,6 @@ You can do this in the command line with `conda activate py4dstem`, or, if you'r
 
 ## Running the GUI
 
-**WARNING: the GUI is currently down. Thanks for you patience as we work through various bugs!**
-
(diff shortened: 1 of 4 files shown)


bsavitzky (approved):
Thanks Steve - as you note, this is a supremely terrible way of reading EMD datacubes, and is also probably the best/only way to do it with the current architecture of the io code - so the right solution here is to refactor that code in a future update.  Let's discuss more soon / post MM.

PR #190, opened 2021-08-31
Initial parallelisation via dask

alex-rakowski:
This PR isn't ready to go into dev, as the docstrings aren't suitable, code needs cleaning, and bug hunting. Apologies the commit history at the end is a bit of a mess. 

This PR address:
- Adding in Dask support for loading out of memory files from py4DSTEM native files. All the py4DSTEM functions I have tried on dask arrays have worked without editing but there is a significant performance penalty. 
- Improving the speed for creating virtual images. Dask guvectorize functions for operating on Dask Arrays, and a numba accelerated function, both of which are parallelized. I couldn't beat the performance of in-memory BF image creation
- Parallelised the disk_detection using dask, to operate on in memory and out of memory datasets. It works on a single machine and in theory, it should scale to HPC systems, but I haven't validated this part yet. 

This framework should also be compatible with #175 and allow for multi GPU clusters


alex-rakowski:
example usage: 
```
dataset = py4DSTEM.io.read(file_path_input,data_id=0, mem="DASK")
from py4DSTEM.process.diskdetection import beta_parallel_disk_detection
new_peaks = beta_parallel_disk_detection(probe=probe_kernel_FT, dataset=dataset, return_dask_client=False)
```

sezelt:
Hey Alex this is quite cool! 

I haven't tested this yet but I will add a few initial thoughts:
- I'm not surprised that Dask slows down most of the py4DSTEM functions. I don't think Dask really has any strategies that would meaningfully accelerate most of the operations we do, so it's just following the same path but with a lot of overhead. This was my experience playing with Dask in py4DSTEM in the past. 
- The guvectorize stuff is new to me, this looks really interesting!
- I am not sure that this approach would scale to a distributed system. I think the secret sauce in Matt Henderson's Dask parallel disk detection code is that the I/O is handled by each worker. From what I understand, in the way you are doing things the head worker will load the data from disk and send it to each worker; this will cause a bottleneck if the workers are on different nodes. 

I'm curious what performance improvement you've seen so far on a workstation. I can test this on our 40-core machine if you'd like...

alex-rakowski:
Hi Steve,

I think I avoided the I/0 bottle neck, each worker is sent a task to which
should involve loading a slice of the datacube. But I didn't check this,
and I could do this more explicitly, I originally had a little function
which loads a h5 slice from a file path and pointer.

I'll check this on nersc later this week, it would be cool to see how fast
the 40 core version goes. I could sent you the notebook I've been using
test the functions.

sezelt:
Aha I see, so the `data.to_delayed()` defers the actual reading to the worker processes?

py4DSTEM/io/native/read/read_v0_12.py
```diff
@@ -93,7 +93,7 @@ def read_v0_12(fp, **kwargs):
         # Parse optional arguments
         if 'mem' in kwargs.keys():
             mem = kwargs['mem']
-            assert(mem in ('RAM','MEMMAP'))
+            assert(mem in ('RAM','MEMMAP', 'DASK'))
```
bsavitzky, line 96: also edit the docstring - line 39, the 'mem' argument - to reflect the 'DASK' option

py4DSTEM/io/read.py
```diff
@@ -72,7 +72,7 @@ def read(fp, mem="RAM", binfactor=1, ft=None, metadata=False, **kwargs):
     """
     assert(isinstance(fp,(str,pathlib.Path))), "Error: filepath fp must be a string or pathlib.Path"
     assert(exists(fp)), "Error: specified filepath does not exist."
-    assert(mem in ['RAM','MEMMAP']), 'Error: argument mem must be either "RAM" or "MEMMAP"'
+    assert(mem in ['RAM','MEMMAP', 'DASK']), 'Error: argument mem must be either "RAM" or "MEMMAP"'
```
bsavitzky, line 75: edit the docstring 'mem' argument description (line 22) to reflect the 'DASK' option

alex-rakowski: Updated the doc string

py4DSTEM/io/native/read/read_v0_12.py
alex-rakowski: Updated the doc string

PR #191, opened 2021-08-31
Transparent masks and other fixes to `show`

sezelt:
This adds the ability to have a partially transparent mask (as well as fixing other bugs in the `show` function related to masking and unmasked NaNs). 

For example:
```python 
noise = np.random.rand(50,50) - 0.5
noise[5:9,5:9] = np.nan

mask = np.zeros_like(noise,dtype=np.bool)
mask[10:40,20:30] = True

cmap='inferno'
```

Plotting just the array with `plt.matshow(noise,cmap=cmap)`:
image

Using py4DSTEM to plot the array with a mask `py4DSTEM.visualize.show(noise,mask=mask,cmap=cmap)`:
image

Using py4DSTEM to display with a partially transparent mask `py4DSTEM.visualize.show(noise,mask=mask,cmap=cmap,mask_alpha=0.2)`:
image

### DIFF
--- a/py4DSTEM/visualize/show.py
+++ b/py4DSTEM/visualize/show.py
@@ -255,6 +255,7 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
         _mask = ar.data>0
         _ar = np.zeros_like(ar.data,dtype=float)
         _ar[_mask] = np.log(ar[_mask])
+        _ar[~_mask] = -np.inf
         if min != None:
             if min > 0:
                 min = np.log(min)
@@ -266,6 +267,7 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
         _mask = ar>0
         _ar = np.zeros_like(ar.data,dtype=float)
         _ar[_mask] = np.power(ar[_mask],power)
+        _ar[~_mask] = -np.inf
         if min != None:
             min = np.power(min,power)
         if max != None:
@@ -310,9 +312,9 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
     # Plot the image
     if not hist:
         cax = ax.matshow(_ar,vmin=vmin,vmax=vmax,cmap=cm,**kwargs)
-        if np.any(_ar.mask==True):
+        if np.any(_ar.mask):
             mask_display = np.ma.array(data=_ar.data,mask=~_ar.mask)
-            ax.matshow(mask_display,cmap=cmap,alpha=mask_alpha)
+            ax.matshow(mask_display,cmap=cmap,alpha=mask_alpha,vmin=vmin,vmax=vmax)
     # ...or, plot its histogram
     else:
         hist,bin_edges = np.histogram(_ar,bins=np.linspace(np.min(_ar),
--- a/py4DSTEM/visualize/show.py
+++ b/py4DSTEM/visualize/show.py
@@ -274,7 +274,8 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
             max = np.power(max,power)
     else:
         raise Exception
-    _ar = np.ma.array(data=_ar.data,mask=np.logical_or(ar.mask,~_mask))
+    
+    _ar = np.ma.array(data=_ar.data,mask=~_mask)
 
     # Set the clipvalues
     if clipvals == 'minmax':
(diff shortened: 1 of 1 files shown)


py4DSTEM/visualize/show.py
```diff
@@ -247,9 +252,10 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
         _ar = ar.copy()
         _mask = np.ones_like(_ar.data,dtype=bool)
     elif scaling == 'log':
-        _mask = ar>0
+        _mask = ar.data>0
         _ar = np.zeros_like(ar.data,dtype=float)
-        _ar[_mask] = np.log(ar[_mask])
+        _ar[_mask] = np.log(ar.data[_mask])
+        _ar[~_mask] = -np.inf
```
bsavitzky, line 258: why -inf and not nan?

py4DSTEM/visualize/show.py
```diff
@@ -258,31 +264,33 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
...
     else:
         raise Exception
-    _ar = np.ma.array(data=_ar.data,mask=np.logical_and(ar.mask==False,_mask)==False)
+    
+    _ar = np.ma.array(data=_ar.data,mask=~_mask)
 
     # Set the clipvalues
     if clipvals == 'minmax':
-        vmin,vmax = np.ma.min(_ar),np.ma.max(_ar)
+        vmin,vmax = np.nanmin(_ar),np.nanmax(_ar)
```
bsavitzky, line 282: do the nanmin/nanmax/nanmedian etc functions play nicely with masked arrays? the old code here could break if there were non-masked nan entries, and this new code will take care of that - however, the old code also ensured we we correctly ignoring masked array elements, which np.min/np.max/np.median/etc don't do.  so, just confirming before pulling this in, will these functions handle both cases?

py4DSTEM/visualize/show.py
```diff
@@ -258,31 +264,33 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
...
     elif clipvals == 'manual':
         assert min is not None and max is not None
         vmin,vmax = min,max
     elif clipvals == 'std':
         assert min is not None and max is not None
-        m,s = np.ma.median(_ar),np.ma.std(_ar)
+        m,s = np.nanmedian(_ar),np.nanstd(_ar)
```
bsavitzky, line 288: see above

py4DSTEM/visualize/show.py
```diff
@@ -258,31 +264,33 @@ def show(ar,figsize=(8,8),cmap='gray',scaling='none',clipvals='minmax',
...
         vmin = m + min*s
         vmax = m + max*s
     elif clipvals == 'centered':
-        c = np.ma.mean(_ar) if min is None else min
-        m = np.ma.max(np.ma.abs(c-_ar)) if max is None else max
+        c = np.nanmean(_ar) if min is None else min
+        m = np.nanmax(np.ma.abs(c-_ar)) if max is None else max
```
bsavitzky, line 293: see above

bsavitzky (changes requested):
looks great, i just had one question about np.nanmin vs np.ma.min

py4DSTEM/visualize/show.py
sezelt: Yes, I was concerned about the same thing so I did check that they respect the mask (why they're different functions, I'm not sure... and there is no such thing as `np.ma.nanmean`, which is confusing) 

The behavior is:
| Data | Mask | `np.mean` | `np.nanmean` | `np.ma.mean` |
| ----- | ------- | ---------| ----------- | ----------- |
| `[1.0, 2.0, nan, nan]` | `[False, False, False, False]` | NaN | 1.5 | 1.5 |
| `[1.0, 2.0, nan, nan]` | `[False, False, False,  True]` | NaN | 1.5 | NaN |
| `[1.0, 2.0, nan, nan]` | `[False, False, True, True]` | 1.5 | 1.5 | 1.5 |
| `[1.0, 2.0, nan, nan]` | `[False,  True, False,  True]` | NaN | 1.0 | NaN |

So `nan___` does indeed respect the mask, and handles any rogue NaNs in the input array

py4DSTEM/visualize/show.py
sezelt: That's... a good question. I think that NaN would work just as well and is probably the more sensible choice. I will change these.

bsavitzky (approved):

PR #203, opened 2021-10-01
Some more small tweaks

sezelt:
This PR does two things:
- Solves the problem in origin.py where testing `isinstance(pla,PointListArray)` would fail even though `pla` was a `PointListArray`. @PhilippPelz pointed us to [this stackoverflow](https://stackoverflow.com/questions/21498211/using-isinstance-in-modules) that explains the problem: importing from `...io.datastructure` would re-import that file, creating a new instance of the `PointListArray` class that is different from the one you call when creating a `PointListArray` through the module. Thus the classes were not the same object, and the `isinstance` test would return `False`. By changing the import to `...io`, we get the same object as the rest of the module, and the problem is solved (at least in this file... similar issues may lurk elsewhere). 
- Transposes the images shown in the GUI so that they match how it would appear in matplotlib or in Digital Micrograph. When using the GUI at the Titan, it is often rather confusing that the image is transposed from how it appeared to the operator.

### DIFF
--- a/py4DSTEM/process/calibration/origin.py
+++ b/py4DSTEM/process/calibration/origin.py
@@ -4,9 +4,9 @@ import numpy as np
 from scipy.ndimage.filters import gaussian_filter
 from scipy.optimize import leastsq
 
-from ..fit import plane,parabola,bezier_two,fit_2D
-from ..utils import get_CoM, add_to_2D_array_from_floats,tqdmnd
-from ...io.datastructure import PointListArray
+from ..fit import plane, parabola, bezier_two, fit_2D
+from ..utils import get_CoM, add_to_2D_array_from_floats, tqdmnd
+from ...io import PointListArray, DataCube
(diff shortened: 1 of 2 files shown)


py4DSTEM/process/calibration/origin.py
```diff
@@ -4,9 +4,9 @@
 from scipy.ndimage.filters import gaussian_filter
 from scipy.optimize import leastsq
 
-from ..fit import plane,parabola,bezier_two,fit_2D
-from ..utils import get_CoM, add_to_2D_array_from_floats,tqdmnd
-from ...io.datastructure import PointListArray
+from ..fit import plane, parabola, bezier_two, fit_2D
+from ..utils import get_CoM, add_to_2D_array_from_floats, tqdmnd
```
bsavitzky, line 8: oh no, autoformatting

bsavitzky (approved):
Looks great, thanks Steve

PR #204, opened 2021-10-04
replaced all "...io.datastructure import" with "...io" 

alex-rakowski:
Ran`grep -rl 'from ...io.datastructure import' | xargs sed -i 's/from ...io.datastructure import/from ...io import/g'`


sezelt (approved):
These changes all look fine to me. I also checked out the branch and py4DSTEM imported successfully, which is probably how any major problems from this change would manifest.

PR #213, opened 2021-10-13
Adding function for arbitrary virtual image generation

cophus:
Copying method used in annular image generation

### DIFF
--- a/py4DSTEM/process/virtualimage/virtualimage.py
+++ b/py4DSTEM/process/virtualimage/virtualimage.py
@@ -86,6 +86,37 @@ def get_virtualimage_ann(datacube, x0, y0, Ri, Ro, verbose=True):
 
 
 
+def get_virtualimage(datacube, mask, verbose=True):
+    """
+    Get a virtual image using an arbitrary boolean mask
+
+    Args:
+        datacube (DataCube):    Input datacube with dimensions (R_Nx, R_Nx, Q_Nx, Q_Ny)
+        mask (bool):            Mask with dimensions (Q_Nx, Q_Ny)
+
+    Returns:
+        (2D array): the virtual image
+    """
+    assert isinstance(datacube, DataCube)
+
+    
+    # assert Ro>Ri, "Inner radius must be smaller than outer radius"
+    # xmin,xmax = max(0,int(np.floor(x0-Ro))),min(datacube.Q_Nx,int(np.ceil(x0+Ro)))
+    # ymin,ymax = max(0,int(np.round(y0-Ro))),min(datacube.Q_Ny,int(np.ceil(y0+Ro)))
+
+    # xsize,ysize = xmax-xmin,ymax-ymin
+    # x0_s,y0_s = x0-xmin,y0-ymin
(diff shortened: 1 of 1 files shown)


cophus:
I placed an example notebook into slack, but the gist of the new function is something like:

mask_inner = py4DSTEM.process.utils.make_circular_mask((dataset.Q_Nx,dataset.Q_Ny), (qx0, qy0), det_DF[0])
mask_outer = py4DSTEM.process.utils.make_circular_mask((dataset.Q_Nx,dataset.Q_Ny), (qx0, qy0), det_DF[1])
mask = mask_outer & ~mask_inner

virtual_image = py4DSTEM.process.virtualimage.get_virtualimage(
    dataset,
    mask)

mask can be any boolean array - this function also uses the square subslicing method for a speed boost.

py4DSTEM/process/virtualimage/virtualimage.py
```diff
@@ -85,7 +85,36 @@ def get_virtualimage_ann(datacube, x0, y0, Ri, Ro, verbose=True):
...
+    x = np.where(np.max(mask, axis=1))
+    y = np.where(np.max(mask, axis=0))
+    xmin, xmax = np.min(x), np.max(x)
+    ymin, ymax = np.min(y), np.max(y)
+    mask_sub = mask[xmin:xmax,ymin:ymax]
+
+    # Generate virtual image
+    virtual_image = np.zeros((datacube.R_Nx, datacube.R_Ny))
+    for rx,ry in tqdmnd(datacube.R_Nx, datacube.R_Ny, disable=not verbose):
+        virtual_image[rx,ry] = np.sum(datacube.data[rx,ry,xmin:xmax,ymin:ymax]*mask_sub)
```
sezelt, line 116: On my machine, this is 13% faster for the annular mask in the example notebook if you replace the multiplication by boolean with slicing by a boolean, i.e.
```
virtual_image[rx,ry] = np.sum(datacube.data[rx,ry,xmin:xmax,ymin:ymax][mask_sub])
```

py4DSTEM/process/virtualimage/virtualimage.py
```diff
@@ -85,7 +85,36 @@ def get_virtualimage_ann(datacube, x0, y0, Ri, Ro, verbose=True):
     return virtual_image
 
 
+def get_virtualimage(
+    datacube, 
+    mask, 
+    verbose=True,
+    return_mask_coords=False,
```
sezelt, line 92: This is unused?

sezelt (changes requested):
Looks great!
There is a small tweak, commented below, that will give a further performance boost (thanks to @PhilippPelz for the suggestion!).
Also, can you update the docstring to reflect the new arguments?

py4DSTEM/process/virtualimage/virtualimage.py
cophus: Oops a leftover from the previous version!

py4DSTEM/process/virtualimage/virtualimage.py
cophus: I always forget you can do this with numpy arrays - added!

cophus: Docstring updated for verbose too

PR #218, opened 2021-10-18
make scaling for radius and intensity user-selectable

sezelt:
This small tweak implements the suggestion that has been floating around about making the scaling powers of both the shell radius and the peak intensity user-selectable. They can be specified at the creation of the orientation plan, and are saved for later use when scaling the correlation images in the pattern matching.

### DIFF
--- a/py4DSTEM/process/diffraction/crystal.py
+++ b/py4DSTEM/process/diffraction/crystal.py
@@ -634,6 +634,8 @@ class Crystal:
         angle_step_in_plane: float = 2.0,
         accel_voltage: float = 300e3,
         corr_kernel_size: float = 0.08,
+        radial_power:float = 1.,
+        intensity_power:float = 0.5,
         tol_peak_delete = None,
         tol_distance: float = 0.01,
         fiber_axis=None,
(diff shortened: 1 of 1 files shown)


py4DSTEM/process/diffraction/crystal.py
```diff
@@ -684,6 +688,10 @@ def orientation_plan(
         # Calculate wavelenth
         self.wavelength = electron_wavelength_angstrom(self.accel_voltage)
 
+        # store the radial and intensity scaling to use later for generating test patterns
+        self.radial_power = radial_power
```
cophus, line 692: Looks good, but to match the convention of the other variables, anything related to orientation mapping is currently stored in the format:  orientation_variable_name.  Therefore I suggest changing these variables to be called:
self.orientation_radial_power = radial_power
self.orientation_intensity_power = intensity_power

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -660,6 +662,8 @@ def orientation_plan(
             angle_step_in_plane (float):  Approximate angular step size for in-plane rotation [degrees]
             accel_voltage (float):        Accelerating voltage for electrons [Volts]
             corr_kernel_size (float):        Correlation kernel size length in Angstroms
+            radial_power (float):          Power for scaling the correlation intensity of each radial bin
+            intensity_power (float):       Power for scaling the intensity of structure factors
```
cophus, line 666: I think these docstrings could be a touch clearer:
Power for scaling the correlation intensity as a function of the peak radius
Power for scaling the correlation intensity as a function of the peak intensity

cophus (changes requested):

py4DSTEM/process/diffraction/crystal.py
sezelt: ✅

PR #229, opened 2021-10-31
small bugfixes for ACOM PR

sezelt:
Somehow, presumably when I was merging things a while ago, some of the functionality of the beamstop center finding function got lost. This PR fixes that error.

There was also a syntax error in crystal.py, where `cartesian_directions` was referenced instead of `self.cartesian_directions`, which I have fixed. 

Running the linter on `crystal.py` also shows an error on line 2513 in `match_single_pattern` when `plot_corr_3D` is True, where an undefined name `corr` is used, but I can't tell by quick inspection what this name is supposed to be referring to.

### DIFF
--- a/py4DSTEM/process/calibration/origin.py
+++ b/py4DSTEM/process/calibration/origin.py
@@ -348,7 +348,7 @@ def get_origin_brightest_disk(
 
     return qx0_ar, qy0_ar
 
-def get_origin_single_dp_beamstop(DP,mask):
+def get_origin_single_dp_beamstop(DP: np.ndarray,mask: np.ndarray):
     """
     Find the origin for a single diffraction pattern, assuming there is a beam stop.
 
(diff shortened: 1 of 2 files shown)


cophus (approved):
Looks good to me - a useful additional function, and a good fix.

cophus:
> Somehow, presumably when I was merging things a while ago, some of the functionality of the beamstop center finding function got lost. This PR fixes that error.
> 
> There was also a syntax error in crystal.py, where `cartesian_directions` was referenced instead of `self.cartesian_directions`, which I have fixed.
> 
> Running the linter on `crystal.py` also shows an error on line 2513 in `match_single_pattern` when `plot_corr_3D` is True, where an undefined name `corr` is used, but I can't tell by quick inspection what this name is supposed to be referring to.

I had a 3D correlation plot - it was removed at some point when I changed the array sizes, and I haven't added it back yet.  This can be removed.

PR #231, opened 2021-11-13
Make BraggVectorMaps ~150x faster

sezelt:
Previously we were calling `add_to_2D_array_from_floats` once _per point_ in making Bragg vector maps! Even though `add_to_2D_array_from_floats` was vectorized!

Anywho, I have replaced those nested for loops with concatenation of all of the PointLists together followed by one single accumulation of the intensity at each coordinate using `np.bincount`. `add_to_2D_array_from_floats` won't work correctly if there is more than one point at the same (qx,qy) location, which was fine when it was called once per PointList (or per point!) but would obviously cause problems when doing all the points at once. 

In my testing this new version is consistently ~150x faster than previously, regardless of the size of the data. For a very large PLA from a 512x512x512x512 scan that I have on hand, the BVM generation time went down from 119s to 0.737s.

This speedup applies to `get_Bragg_vector_map` and `get_bragg_vector_map_raw` only. Initially I had reworked `add_to_2D_array_from_floats` using the `np.bincount` based implementation, but for very small inputs (a few points), this could be substantially slower than previously! So instead I have just removed the inner for loop for a ~10x speedup on the weighted BVM functions. One could also rework those to include the weights in the concatenation step, somehow, for the full speedup.

The maximal BVM functions are untouched.

### DIFF
--- a/py4DSTEM/process/diskdetection/braggvectormap.py
+++ b/py4DSTEM/process/diskdetection/braggvectormap.py
@@ -32,7 +32,7 @@ def get_bragg_vector_map(braggpeaks, Q_Nx, Q_Ny):
     ceily = np.ceil(qy).astype(np.int64)
     
     # Remove any points outside [0, Q_Nx] & [0, Q_Ny]
-    mask = np.logical_and(np.logical_and(np.logical_and((floorx>=0),(floory>=0)),(ceilx<Q_Nx)),(ceily<Q_Ny))
+    mask = np.logical_and.reduce(((floorx>=0),(floory>=0),(ceilx<Q_Nx),(ceily<Q_Ny)))
     qx = qx[mask]
     qy = qy[mask]
     I = I[mask]
@@ -152,7 +152,7 @@ def get_bragg_vector_map_raw(braggpeaks, Q_Nx, Q_Ny):
     ceily = np.ceil(qy).astype(np.int64)
     
     # Remove any points outside [0, Q_Nx] & [0, Q_Ny]
-    mask = np.logical_and(np.logical_and(np.logical_and((floorx>=0),(floory>=0)),(ceilx<Q_Nx)),(ceily<Q_Ny))
+    mask = np.logical_and.reduce(((floorx>=0),(floory>=0),(ceilx<Q_Nx),(ceily<Q_Ny)))
     qx = qx[mask]
     qy = qy[mask]
     I = I[mask]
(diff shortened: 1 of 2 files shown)


py4DSTEM/process/diskdetection/braggvectormap.py
```diff
@@ -19,17 +19,48 @@ def get_bragg_vector_map(braggpeaks, Q_Nx, Q_Ny):
...
+    I = bigpl['intensity']
+    
+    # Precompute rounded coordinates
+    floorx = np.floor(qx).astype(np.int64)
+    ceilx = np.ceil(qx).astype(np.int64)
+    floory = np.floor(qy).astype(np.int64)
+    ceily = np.ceil(qy).astype(np.int64)
+    
+    # Remove any points outside [0, Q_Nx] & [0, Q_Ny]
+    mask = np.logical_and(np.logical_and(np.logical_and((floorx>=0),(floory>=0)),(ceilx<Q_Nx)),(ceily<Q_Ny))
```
cophus, line 35: Is there a better way to write this line with .reduce?  For example:

`mask = np.logical_ad.reduce( (floorx>=0),(floory>=0),(ceilx<Q_Nx),(ceily<Q_Ny) )`

py4DSTEM/process/diskdetection/braggvectormap.py
```diff
@@ -109,16 +139,48 @@ def get_bragg_vector_map_raw(braggpeaks, Q_Nx, Q_Ny):
...
```
cophus, line 155: Same commend as above, i.e.:
`mask = np.logical_ad.reduce( (floorx>=0),(floory>=0),(ceilx<Q_Nx),(ceily<Q_Ny) )`
?

cophus (changes requested):

py4DSTEM/process/diskdetection/braggvectormap.py
sezelt: In quick testing, using `logical_and.reduce` is actually about 20% slower for this application! But it's far less than 1% of the total time of the function, and it looks neater, so I'm going to change it!

py4DSTEM/process/diskdetection/braggvectormap.py
sezelt: same as above

bsavitzky (approved):

py4DSTEM/process/diskdetection/braggvectormap.py
bsavitzky: these have been changed, resolving the convos

bsavitzky:
Merged with @cophus's review outstanding b/c the edits had been made, just hadn't been formally approved in his review yet - Colin give a shout if there was anything missing that we should add

PR #232, opened 2021-11-18
Adding pre-processing functions for ML processing

cophus:
Updated 2D image shift to include "bilinear" option, and non-periodic boundary functions.  Utility function to apply it to a DataCube.  New pixel filtering function that operates on a DataCube.

### DIFF
--- a/py4DSTEM/process/preprocess/preprocess.py
+++ b/py4DSTEM/process/preprocess/preprocess.py
@@ -150,21 +150,23 @@ def bin_data_real(datacube, bin_factor):
         datacube.R_Nx,datacube.R_Ny,datacube.Q_Nx,datacube.Q_Ny = datacube.data.shape
         return datacube
 
-def filter_bad_pixels(
+def filter_hot_pixels(
     datacube,
     thresh,
     ind_compare=1,
     return_mask=False
     ):
     """
-    This function performs pixel filtering to remove hot or cold pixels. We first compute a moving median filter,
-    applied to the mean diffraction image. Pixels with intensity int_0 are compared to a locally ordered median
-    value, default is the 2nd brightest pixel.
+    This function performs pixel filtering to remove hot / bright pixels. We first compute a moving local ordering filter,
+    applied to the mean diffraction image. This ordering filter will return a single value from the local sorted intensity 
+    values, given by ind_compare. ind_compare=0 would be the highest intensity, =1 would be the second hightest, etc.
+    Next, a mask is generated for all pixels which are least a value thresh higher than the local ordering filter output.
+    Finally, we loop through all diffraction images, and any pixels defined by mask are replaced by their 3x3 local median.
 
     Args:
-            datacube (DataCube):      
-            thresh (float):            threshold for mask
-            ind_compare (int):         which median filter value to compare against. 0 = brightest pixel, 1 = next brightest, etc.
+            datacube (DataCube):      py4DSTEM Datacube
+            thresh (float):           threshold for replacing hot pixels, if pixel value minus local ordering filter exceeds it.
+            ind_compare (int):        which median filter value to compare against. 0 = brightest pixel, 1 = next brightest, etc.
 
         Returns:
             datacube                     datacube              
@@ -227,19 +229,19 @@ def datacube_diffraction_shift(
     bilinear=False,
     ):
     """
-    This function performs pixel filtering to remove hot or cold pixels. We first compute a moving median filter,
-    applied to the mean diffraction image. Pixels with intensity int_0 are compared to a locally ordered median
-    value, default is the 2nd brightest pixel.
+    This function shifts each 2D diffraction image by the values defined by (xshifts,yshifts). 
+    The shift values can be scalars (same shift for all images) or arrays with the same dimensions as
+    the probe positions in datacube.
 
     Args:
-            datacube (DataCube):      
-            xshifts (float):       array or scalar value for the x dim shifts
-            yshifts (float):       array or scalar value for the y dim shifts
-            periodic (bool):       flag for periodic boundary conditions
-            bilinear (bool):       flag for bilinear image shifts
+            datacube (DataCube):   py4DSTEM DataCube
+            xshifts (float):       Array or scalar value for the x dim shifts
+            yshifts (float):       Array or scalar value for the y dim shifts
+            periodic (bool):       Flag for periodic boundary conditions. If set to false, boundaries are assumed to be periodic.
+            bilinear (bool):       Flag for bilinear image shifts. If set to False, Fourier shifting is used.
 
         Returns:
-            datacube                     datacube              
+            datacube (DataCube):   py4DSTEM DataCube              
     """
 
     # if the shift values are constant, expand to arrays
@@ -261,3 +263,4 @@ def datacube_diffraction_shift(
         )
 
     return datacube
+    
--- a/py4DSTEM/process/preprocess/preprocess.py
+++ b/py4DSTEM/process/preprocess/preprocess.py
@@ -153,11 +153,21 @@ def filter_bad_pixels(
     datacube,
     thresh,
     ind_compare=1,
+    return_mask=False
     ):
     """
     This function performs pixel filtering to remove hot or cold pixels. We first compute a moving median filter,
     applied to the mean diffraction image. Pixels with intensity int_0 are compared to a locally ordered median
     value, default is the 2nd brightest pixel.
+
+    Args:
+            datacube (DataCube):      
+            thresh (float):            threshold for mask
+            ind_compare (int):         which median filter value to compare against. 0 = brightest pixel, 1 = next brightest, etc.
+
+        Returns:
+            datacube                     datacube              
+            mask (bool):                 (optional) the bad pixel mask
     """
 
     # Mean image over all probe positions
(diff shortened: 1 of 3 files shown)


py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,114 @@ def bin_data_real(datacube, bin_factor):
...
+    datacube,
+    xshifts,
+    yshifts,
+    periodic=True,
+    bilinear=False,
+    ):
+    """
+    This function performs pixel filtering to remove hot or cold pixels. We first compute a moving median filter,
+    applied to the mean diffraction image. Pixels with intensity int_0 are compared to a locally ordered median
+    value, default is the 2nd brightest pixel.
```
sezelt, line 232: This appears to be the wrong docstring

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,114 @@ def bin_data_real(datacube, bin_factor):
...
+
+    Args:
+            datacube (DataCube):      
+            xshifts (float):       array or scalar value for the x dim shifts
+            yshifts (float):       array or scalar value for the y dim shifts
+            periodic (bool):       flag for periodic boundary conditions
+            bilinear (bool):       flag for bilinear image shifts
```
sezelt, line 239: Maybe note that if bilinear is off, Fourier is used

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,114 @@ def bin_data_real(datacube, bin_factor):
...
+    for ax, ay in tqdmnd(*(datacube.R_Nx,datacube.R_Ny), desc="Shifting images", unit=" images"):
+        datacube.data[ax,ay,:,:] = get_shifted_ar(
+            datacube.data[ax,ay,:,:],
+            xshifts[ax,ay],
+            yshifts[ax,ay],
+            periodic=periodic,
+            bilinear=bilinear,
+        )
 
+    return datacube
```
sezelt, line 263: Missing newline at end of file

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,114 @@ def bin_data_real(datacube, bin_factor):
...
 
+def filter_bad_pixels(
+    datacube,
+    thresh,
+    ind_compare=1,
+    return_mask=False
+    ):
+    """
+    This function performs pixel filtering to remove hot or cold pixels. We first compute a moving median filter,
+    applied to the mean diffraction image. Pixels with intensity int_0 are compared to a locally ordered median
```
sezelt, line 161: Can you explain what `thresh` does in the docstring? It's unclear. Something more like "pixels are compared to their 21 neighbors, and those with intensity greater than the `ind_compare`th brightest neighbor by `thresh` are marked as "hot" and replaced with the value of the 8-neighbor median in each diffraction pattern".

Yes, writing that out made me feel insane. But I feel we should describe it precisely. 

Also, does this actually detect cold pixels? I think it is only doing ones that are stuck high.

sezelt (changes requested):
My only comments are about the docstrings and function names. One function has the wrong docstring.

py4DSTEM/process/preprocess/preprocess.py
cophus: Doc string expanded, function renamed to specify hot pixels.

py4DSTEM/process/preprocess/preprocess.py
cophus: done

py4DSTEM/process/preprocess/preprocess.py
cophus: Good catch!  Lazy copy and pasting on my part.

sezelt (approved):
Looks good to me

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,117 @@ def bin_data_real(datacube, bin_factor):
...
+            thresh (float):           threshold for replacing hot pixels, if pixel value minus local ordering filter exceeds it.
+            ind_compare (int):        which median filter value to compare against. 0 = brightest pixel, 1 = next brightest, etc.
+
+        Returns:
+            datacube                     datacube              
+            mask (bool):                 (optional) the bad pixel mask
+    """
+
+    # Mean image over all probe positions
+    diff_mean = np.mean(datacube.data,axis=(0,1))
```
bsavitzky, line 177: in the future we might have this call a datacube fn which (a) looks to see if this has already been computed and if so returns it, else (b) computes and stores it

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,117 @@ def bin_data_real(datacube, bin_factor):
...
+
+    # Local median
```
bsavitzky, line 179: I'd suggest expanding this comment - we're not just computing the local median - if we were we could just call the scipy fn we've imported! - we're actually getting a sorted array of local intensities at each pixel, right?

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,117 @@ def bin_data_real(datacube, bin_factor):
...
+    # Generate mask
+    mask = diff_mean - diff_compare > thresh
+
+    # apply filtering
+    for ax, ay in tqdmnd(*(datacube.R_Nx,datacube.R_Ny), desc="Cleaning pixels", unit=" images"):
+        # Calculate local 3x3 median images
+        im_med = median_filter(
+            datacube.data[ax,ay,:,:],
+            size=3,
+            mode='nearest')
```
bsavitzky, line 216: for consistency with the `diff_compare` array, I would suggest changing to `mode='wrap'`

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,117 @@ def bin_data_real(datacube, bin_factor):
...
+            datacube (DataCube):   py4DSTEM DataCube              
+    """
+
+    # if the shift values are constant, expand to arrays
+    xshifts = np.array(xshifts)
+    yshifts = np.array(yshifts)
+    if xshifts.ndim == 0:
+        xshifts = xshifts * np.ones((datacube.R_Nx,datacube.R_Ny))
+    if yshifts.ndim == 0:
+        yshifts = yshifts * np.ones((datacube.R_Nx,datacube.R_Ny))
```
bsavitzky, line 253: this should not be necessary - numpy will cast these automatically

py4DSTEM/process/preprocess/preprocess.py
```diff
@@ -149,5 +150,117 @@ def bin_data_real(datacube, bin_factor):
...
+    # if the shift values are constant, expand to arrays
+    xshifts = np.array(xshifts)
+    yshifts = np.array(yshifts)
+    if xshifts.ndim == 0:
+        xshifts = xshifts * np.ones((datacube.R_Nx,datacube.R_Ny))
+    if yshifts.ndim == 0:
+        yshifts = yshifts * np.ones((datacube.R_Nx,datacube.R_Ny))    
+
+    # Loop over all images
+    for ax, ay in tqdmnd(*(datacube.R_Nx,datacube.R_Ny), desc="Shifting images", unit=" images"):
```
bsavitzky, line 256: I'm a little confused why the `*` operator is needed here? maybe we should add a docstring to the tqdmnd function

py4DSTEM/process/utils/utils.py
```diff
@@ -152,17 +152,65 @@ def get_shift(ar1, ar2, corrPower=1):
...
+def get_shifted_ar(
+    ar, 
+    xshift, 
+    yshift,
+    periodic=True,
+    bilinear=False,
+    ):
+    """
+	Shifts array ar by the shift vector (xshift,yshift), using the either 
+    the Fourier shift theorem (i.e.	with sinc interpolation), or bilinear
```
bsavitzky, line 164: rm random whitespace

py4DSTEM/process/utils/utils.py
```diff
@@ -152,17 +152,65 @@ def get_shift(ar1, ar2, corrPower=1):
...
+    Args:
+            ar (float):            input array 
+            xshift (float):        shift along axis 0 (x) in pixels
+            yshift (float):        shift along axis 1 (y) in pixels
+            periodic (bool):       flag for periodic boundary conditions
+            bilinear (bool):       flag for bilinear image shifts
+
+        Returns:
+            datacube                     datacube              
+            mask (bool):                 (optional) the bad pixel mask
```
bsavitzky, line 176: this doesn't appear to be an actual output, we should remove it

py4DSTEM/process/utils/utils.py
```diff
@@ -152,17 +152,65 @@ def get_shift(ar1, ar2, corrPower=1):
...
+        yR = (np.round(yshift)).astype(int)
+
+        if xR > 0:
+            shifted_ar[0:xR,:] = 0
+        elif xR < 0:
+            shifted_ar[xR:,:] = 0    
+        if yR > 0:
+            shifted_ar[:,0:yR] = 0
+        elif yR < 0:
+            shifted_ar[:,yR:] = 0
```
bsavitzky, line 212: we might consider changing this to apply the same bounary behavoir as, e.g. `scipy.ndimage.median_filter(..., mode='nearest')`, rather than just applying zeros.  Probably not worth the effort without a real reason to though

bsavitzky (approved):
Looks great to me, a few suggestions in the comments, none of them urgent

py4DSTEM/process/preprocess/preprocess.py
cophus: True - originally it was going to be a median filter.  The top level docstring spells it out, but we could extend the definition here too.

py4DSTEM/process/preprocess/preprocess.py
cophus: What would we use wrap?  I think diff_compare should be changed!  The diffraction images are not periodic right?

py4DSTEM/process/preprocess/preprocess.py
cophus: How?  If we pass an int and then index:
`xshifts[ax,ay] `
won't it crash?

py4DSTEM/process/preprocess/preprocess.py
cophus: I don't know what the * does, I copied it from some code from @sezelt

py4DSTEM/process/utils/utils.py
cophus: I think "nearest" is incorrect here, and zeros are more likely to be correct.  "nearest" would turn a disk on the edge of the image into a long streak.

py4DSTEM/process/utils/utils.py
cophus: It's possible I forgot to push that change, but I am pretty sure it's in there now . . .

py4DSTEM/process/preprocess/preprocess.py
bsavitzky: Changing `diff_compare` would also work - the main thing I want is for them to use the same boundary handling, whatever it is

py4DSTEM/process/preprocess/preprocess.py
bsavitzky: oops! yup you are right, i'm being silly

py4DSTEM/process/preprocess/preprocess.py
bsavitzky: meh, ok, good enough for now.  the * should unpack the tuple, so that the `tqdmnd` function receives each of those arguments within that tuple individually, i think.  but i don't understand why this is what we're doing, b/c tqdmnd has no documentation, so i'm not really sure what it expect.  so that's mostly what i was complaining about.....not actually related to this PR!

py4DSTEM/process/utils/utils.py
bsavitzky: Nope, it isn't!

py4DSTEM/process/utils/utils.py
bsavitzky: Ok, yup, agreed

py4DSTEM/process/preprocess/preprocess.py
bsavitzky: fixed

py4DSTEM/process/utils/utils.py
bsavitzky: fixed

py4DSTEM/process/utils/utils.py
bsavitzky: fixed

py4DSTEM/process/preprocess/preprocess.py
bsavitzky: This is honestly a pretty minor point, I don't expect it will make much difference.  Using `mode='constant',cval=0` might make the most sense, given that these are most likely high |k| points in diffraction space....but I won't fiddle, this seems fine

PR #247, opened 2022-02-11
Bloch wave

sezelt:
This PR adds dynamical diffraction simulations to the `Crystal` class, using the Bloch wave method.

Specifically, using this code one can generate:
* Plane wave diffraction patterns, stored in a `PointList`
* Converged beam diffraction patterns (including position-averaged CBED), stored in a numpy array
* Large-angle CBED, stored in a dictionary of numpy arrays, indexed by the (h,k,l) for each reflection

Each above functionality also has the option to efficiently compute a thickness series of patterns, which are returned as lists of the above datastructures.

The basic workflow is:
* Generate a `Crystal` object
* Compute the relativistic corrected potential coefficients, optionally including phonon and plasmon absorption
* Generate a list of beams to include in the Bloch calculation using the kinematic diffraction pattern generator
* Perform the dynamical calculation using those beams and the pre-computed relativistic absorptive potentials

This PR completely restructures the `py4DSTEM.process.diffraction` section of the code, since it adds so much new material that the old omnibus `crystal.py` would have become a 10,000-line affair. Thus `Crystal` is now split over several files, but still behaves as a single class.

There are very few changes outside of the `diffraction` submodule:
* Some behavior of the `PointList` initialization is optimized for performance
* `DataObject` tracking is turned off by default, since this was causing enormous performance problems and insane memory usage now that we have started using `PointLists` as throwaway objects for internal purposes. 
* Some unit errors are corrected in `single_atom_scatter.py` when using `'VA'` return mode.

The attached notebook demonstrates most of the functionality of the code: 
[bloch wave PR demo final.ipynb.zip](https://github.com/py4dstem/py4DSTEM/files/8051594/bloch.wave.PR.demo.final.ipynb.zip)


bsavitzky (approved):
Let the era of the Bloch waves BEGIN

PR #250, opened 2022-02-18
ACOM update

cophus:
This update refactors ACOM to use dataclasses for both orientations and orientation_maps.    This allows easy access to parameters such as Euler angles, and is a necessary step for making colored orientation maps for all  symmetry groups (will require a pymatgen update).


py4DSTEM/process/diffraction/utils.py
```diff
@@ -0,0 +1,100 @@
...
+    angles = None
+
+    def __post_init__(self):
+        # initialize empty arrays
+        self.matrix = np.zeros((self.num_matches,3,3))
+        self.family = np.zeros((self.num_matches,3,3))
+        self.corr   = np.zeros((self.num_matches))
+        self.inds   = np.zeros((self.num_matches,2), dtype='int')
+        self.mirror = np.zeros((self.num_matches), dtype='bool')
+        self.angles = np.zeros((self.num_matches,3))
```
alex-rakowski, line 28: Did @cophus  or @sezelt check if this is the correct way to initialize? I remember Steve mentioning this may create objects that all point at the same mem id?

bsavitzky: This is not correct!  I can fix it

sezelt: Indeed it needs to use a regular `__init__`. If you use the `__post_init__`, then you can't initialize it using a regular constructor and passing in arrays (i.e. `orient = Orientation(matrix=my_array)` would not actually copy in `my_array`), and if you use the default value syntax then all instances would share the same array (i.e. `a = Orientation(); b = Orientation(); a.matrix is b.matrix # True`).

bsavitzky: Two issues:

(1) the type hinting is what @dataclass uses to autogenerate and __init__() function for the underlying class.  So everything after num_matrix is not really doing anything (or, actually, is defining a global variable that exists in the namespace of the Class but not in specific class instances, I believe).  So this needs to be changed to, e.g.

`matrix: np.ndarray = None`

(2) Even if that was done correctly, __post_init__ is overwriting all those values.  So any value a use inputs for `matrix` is being overwritten.

So @cophus - the one question I have is, do you want these initialized by default to `None` or `np.zeros((self.num_matches,3,3))`?  The latter, yes?

py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -1,10 +1,17 @@
...
 
 from ...io.datastructure import PointList, PointListArray
 from ..utils import tqdmnd, electron_wavelength_angstrom
+from .utils import Orientation, OrientationMap, axisEqual3D
 
+try:
+    from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
+    from pymatgen.core.structure import Structure
+except ImportError:
+    pass
```
alex-rakowski, line 14: This will fail silently, should we raise a warning about some features not being available?

py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -76,65 +88,98 @@ def orientation_plan(
...
+        # initialize structure
         structure = Structure(
             self.lat_real, self.numbers, self.positions, coords_are_cartesian=False
         )
 
-        pointgroup = SpacegroupAnalyzer(structure).get_point_group_symbol()
-        self.pointgroup = pointgroup
+        # pointgroup = SpacegroupAnalyzer(structure).get_point_group_symbol()
+        # self.pointgroup = pointgroup
+        self.pointgroup = SpacegroupAnalyzer(structure)
```
alex-rakowski, line 101: Should this be in a try statement, as it is liked to pymatgen? If it fails it could raise a warning that pymatgen is required.

py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -76,65 +88,98 @@ def orientation_plan(
...
         ), "Unrecognized pointgroup returned by pymatgen!"
 
-        zone_axis_range, fiber_axis, fiber_angles = orientation_ranges[pointgroup]
+        zone_axis_range, fiber_axis, fiber_angles = orientation_ranges[ \
+            self.pointgroup.get_point_group_symbol()]
         if isinstance(zone_axis_range, list):
             zone_axis_range = np.array(zone_axis_range)
         elif zone_axis_range == "fiber":
             self.orientation_fiber_axis = np.asarray(fiber_axis)
             self.orientation_fiber_angles = np.asarray(fiber_angles)
```
alex-rakowski, line 113: isn't this already set on line ~70

py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -76,65 +88,98 @@ def orientation_plan(
...
-        )
 
         print(
-            f"Automatically detected point group {pointgroup}, using arguments: zone_axis_range={zone_axis_range}, fiber_axis={fiber_axis}, fiber_angles={fiber_angles}."
+            f"Automatically detected point group {self.pointgroup.get_point_group_symbol()},\n"
+            f" using arguments: zone_axis_range = \n{zone_axis_range}, \n fiber_axis={fiber_axis}, fiber_angles={fiber_angles}."
         )
 
+        # Set a flag so we know pymatgen is available
+        self.pymatgen_available = True
```
alex-rakowski, line 121: A few issues:

1. this is hard codded, and I can't see any logic as to how it could be set to False. Could do something like ~ `self.pymatgen_available = "pymatgen" in sys.modules` earlier on which will dynamically set the flag 

2. pymatgen functionality can be required on line 94  before this flag is set

py4DSTEM/process/diffraction/crystal_ACOM.py
alex-rakowski: or it could be an if statement if the `pymatgen_available` flag is set before this.

bsavitzky (approved):
I'm going to merge this code to keep things moving.  That said, I made quite a few comments.  Eventually it would be great to have them all addressed.  Let's see what can be worked out in threads here and by continuing to add new commits, and then we can discuss whatever remains in an upcoming dev team meeting.

I did I fairly detailed pass through `crystal.py` and `utils.py` (which should be renamed!). Please see comments in `crystal.py`.  In `utils.py` the `@dataclass` structure was not being used quite right such that the behavior may or may not be what you were intending!  I'll make some quick fixes to this and commit them before merging this momentarily.

I went briefly through updates to `flowlines.py`, as these seem to be mostly compatibility edits to make these two modules talk nicely and understand the new datastructures correctly. 

I also made a relatively shallow pass through `crystal_ACOM.py`.  My main comments here are that the pymatgen dependency isn't handled so clearly, and that there is a lot of visualization code mixed with functional code, which should be moved to `crystal_viz.py`. I also noticed at this point that the acceleration voltage is being set in several different places, including here and in `crystal.py`, and that this is probably a little dangerous.

I only skimmed through `crystal_viz.py` for now.

py4DSTEM/process/diffraction/utils.py
sezelt: @bsavitzky the test case you wrote here is not the problem with this syntax. When Python evaluates the class definition, it evaluates the expression in the default assignment, and uses that result for every instantiation of the class. In your example, `Test.a` and `Test.b` are indeed distinct expressions and so yield distinct objects. However, the problem is that _every instance of the dataclass gets the same object_. `Test.a` and `Test.b` are different objects, but every time you create a new `Test`, those will all get _the same `a` and `b` arrays_. Example:
```
In [4]: @dataclass
   ...: class Test:
   ...:     a: np.ndarray = np.zeros((2,2))
   ...:     b: np.ndarray = np.zeros((2,2))
   ...:

In [5]: A = Test()

In [6]: B = Test()

In [7]: A.a is B.a
Out[7]: True

In [8]: A.a[0,0] = 42

In [9]: B.a
Out[9]:
array([[42.,  0.],
       [ 0.,  0.]])
``` 

So the course of action is not the same–to get the behavior that we are expecting you _need_ to use an `__init__`, because it gets evaluated every time the class gets instantiated and thus yields unique objects!

(P.S. I am not actually familiar with the behind-the-scenes mechanics of `dataclass`es, so there may be a way to defer evaluation of the defaults, so that they get reëvaluated every time it gets instantiated. Or it may be the case that using an `__init__` is just how you accomplish this...)

(P.P.S. Python does not do anything with the type hints other than evaluate them to check that they are valid expressions, they can literally be anything that doesn't yield an error. They do not enforce any type discipline and do not effect the resulting objects.)

(P.P.P.S Yes, I did use a New Yorker-style guide diaresis in the first postscript.)

sezelt: Cool!

py4DSTEM/process/diffraction/crystal_ACOM.py
cophus: Yes we should drop a warning - I think it's this way so you can still run everything without pymatgen.

@bsavitzky we could just add pymatgen as a dependency, it seems quite lightweight . . .

PR #256, opened 2022-03-24
Batch FFTs in CUDA disk detection

sezelt:
This adds an optimization to the CUDA Bragg disk detection routine, which is to perform the FFT and probe kernel multiplication step on several diffraction patterns at once on the GPU to better utilize resources. This gives approximately a 60% speedup over the previous CUDA method on my RTX3090.

With a 112x160x864x864 dataset, the performance on my workstation is:
* CPU only: 7.80 DP/s
* CUDA (previous): 66.4 DP/s
* CUDA (with batching): 110.3 DP/s

I have checked that it gives the same results as the old version, and as the CPU version. 

There is also a tweak to the K2 data reader, which theoretically allows it to handle data files that do not begin with a sync byte, which I include here half because it's so small of a change that it doesn't warrant its own PR and half because I forgot it's in my dev branch.

### DIFF
--- a/py4DSTEM/io/nonnative/read_K2.py
+++ b/py4DSTEM/io/nonnative/read_K2.py
@@ -274,11 +274,33 @@ class K2DataArray(Sequence):
         self._bin_files = np.empty(8, dtype=object)
         for i in range(8):
             binName = self._bin_prefix + str(i + 1) + ".bin"
+
+            # Synchronize to the magic sync word
+            # First, open the file in binary mode and read ~1 MB
+            with open(binName, 'rb') as f:
+                s = f.read(1_000_000)
+
+            # Scan the chunk and find everywhere the sync word appears
+            sync = [s.find(b'\xff\xff\x00\x55'),]
+            while sync[-1] >= 0:
+                sync.append(s.find(b'\xff\xff\x00\x55',sync[-1]+1))
+
+            # Since the sync word can conceivably occur within the data region,
+            # check that there is another sync word 22360 bytes away
+            sync_idx = 0
+            while 0 not in [s - sync[sync_idx] - 22360 for s in sync]:
+                sync_idx += 1
+
+            if sync_idx > 0:
+                print(f"Beginning file {i} at offset {sync[sync_idx]} due to incomplete data block!")
+
+            # Start the memmap at the offset of the sync byte
             self._bin_files[i] = np.memmap(
                 binName,
                 dtype=self._stripe_dtype,
                 mode="r",
                 shape=(self._guess_number_frames(),),
+                offset=sync[sync_idx],
             )
 
     def _find_offsets(self):
--- a/py4DSTEM/process/diskdetection/diskdetection_cuda.py
+++ b/py4DSTEM/process/diskdetection/diskdetection_cuda.py
@@ -130,11 +130,13 @@ def find_Bragg_disks_CUDA(
     t0 = time()
     if batching:
         # estimate the max batch size: (this is probably wrong)
-        max_num_bytes = cp.get_default_memory_pool().get_limit()
-        num_batches = (bytes_per_pattern * 3) // max_num_bytes + 1 # what should this fudge factor be?
-        batch_size = datacube.R_N // num_batches + 1
+        max_num_bytes = cp.cuda.Device().mem_info[0]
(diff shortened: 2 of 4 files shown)


py4DSTEM/process/latticevectors/index.py
```diff
@@ -201,22 +202,21 @@ def add_indices_to_braggpeaks(braggpeaks, lattice, maxPeakSpacing, qx_shift=0,
...
+            for i in range(pl.length):
+                r2 = (pl.data['qx'][i]-lattice.data['qx'] + qx_shift)**2 + \
+                     (pl.data['qy'][i]-lattice.data['qy'] + qy_shift)**2
+                ind = np.argmin(r2)
+                if r2[ind] <= maxPeakSpacing**2:
+                    pl.data['h'][i] = lattice.data['h'][ind]
+                    pl.data['k'][i] = lattice.data['k'][ind]
+                else:
+                    rm_peak_mask[i] = True
+            pl.remove_points(rm_peak_mask)
```
alex-rakowski, line 219: Just to double-check you've just replaced a double for loop with a single for loop using tqdmnd?

py4DSTEM/io/nonnative/read_K2.py
```diff
@@ -274,11 +274,33 @@ def _attach_to_files(self):
...
+            # Since the sync word can conceivably occur within the data region,
+            # check that there is another sync word 22360 bytes away
+            sync_idx = 0
+            while 0 not in [s - sync[sync_idx] - 22360 for s in sync]:
+                sync_idx += 1
+
+            if sync_idx > 0:
+                print(f"Beginning file {i} at offset {sync[sync_idx]} due to incomplete data block!")
+
+            # Start the memmap at the offset of the sync byte
```
alex-rakowski, line 297: I don't know enough about the file format to comment, but it looks error free to me.

py4DSTEM/process/latticevectors/index.py
sezelt: Yes, this function was not instantaneous so I decided it would be good to add a progress bar

py4DSTEM/io/nonnative/read_K2.py
sezelt: I've tested it to at least still successfully open files that don't have the specific corruption it addresses, so it's at least not a regression

py4DSTEM/process/diskdetection/diskdetection_cuda.py
```diff
@@ -124,43 +128,116 @@ def find_Bragg_disks_CUDA(
...
-            blocks=blocks,
-            threads=threads,
-        )
+    if batching:
+        # compute the batch size based on available VRAM:
+        max_num_bytes = cp.cuda.Device().mem_info[0]
+        # use a fudge factor to leave room for the fourier transformed data
+        # I have set this at 10, which results in underutilization of 
+        # VRAM, because this yielded better performance in my testing
+        batch_size = max_num_bytes // (bytes_per_pattern * 10)
```
alex-rakowski, line 138: what other values did you try for this? Did you try datasets with different types and/or sized diffraction patterns? I wonder if there's way to pick this value dynamically? But seems sensible for now.

alex-rakowski (commented):
Everything looks good to me.  Have you tried running on multiple datasets with different diffraction pattern sizes and/or data types? 

I can't comment on `read_K2.py` other than it looks error free.

py4DSTEM/io/nonnative/read_K2.py
alex-rakowski: Sounds good to me, I guess we'll have to wait for a corrupted file to confirm it works.

py4DSTEM/process/diskdetection/diskdetection_cuda.py
sezelt: A value of 5 will cause to VRAM to be nearly completely full, which seems to either cause the garbage collector to get invoked more often or to limit the available working RAM for the FFT. So below ~5 it just won't run at all, at 5 it runs perhaps 5% slower than at 10. If you keep adjusting it up, at some point you won't be delivering enough work to the GPU to saturate it at each FFT step, but already the FFT isn't the bottleneck anyway. So this number isn't critical

alex-rakowski (approved):

PR #260, opened 2022-04-05
ACOM patch

sezelt:
In the `Orientation` dataclass, the `__post_init__` function did not actually save the arrays to the class! This PR adds `self.` in the right places.


cophus:
Oops - looks good!

cophus (approved):

PR #261, opened 2022-04-06
Fixing orientation correlation bug

cophus:
I accidentally took the Fourier transform before padding, which means the padding was performing Fourier interpolation rather than padding the array in the image plane.  Order of operations bug!   Should be fixed now, and I also tested normalization and it looks correct.


bsavitzky (approved):
Thanks Colin.  Added in one other small bugfix.

PR #266, opened 2022-04-22
Use float32 in CUDA disk detection

sezelt:
This switches the CUDA disk detection to work in 32-bit floats (or 64-bit complex), which in theory improves performance (depending on the type of GPU you have) at no noticeable accuracy cost.


cophus (approved):
Seems pretty straightforward!   Good find.

PR #268, opened 2022-04-26
Update README.md

alex-rakowski:


README.md
```diff
@@ -67,20 +67,16 @@ If you are running on Linux/Unix machine with Nvidia GPU and CUDA capability, ru
...
 If you are running on Windows with Nvidia GPU and CUDA capability, run
 ```
 conda update conda
-conda create -n py4dstem python==3.8
-conda activate py4dstem
-conda install pip
-pip install tensorflow-gpu==2.4.1
+conda create -n py4dstem-aiml python=3.8 -y && conda activate py4dstem-aiml
+conda install -c conda-forge cudatoolkit=11.0 cudnn=8.1 cupy 
 pip install py4dstem[aiml-cuda]
```
sezelt, line 80: One thing that has tripped up me and Colin is that if you use `zsh` the square brackets are special characters, so it will give a cryptic error unless you escape the string:
`pip install "py4DSTEM[aiml-cuda]"`

alex-rakowski: just checked and `pip install "py4DSTEM[aiml-cuda]"` is safe to use for `bash`, so I'll update to this

sezelt (approved):

bsavitzky:
This looks good - though, I think the plan had been to wait on merging until @cophus had had a chance to test on his windows machine.  As long as that does end up working I think we should be all set with this edit.  Thank @alex-rakowski, @sezelt .

PR #270, opened 2022-05-05
Assorted little fixes

sezelt:
This PR addresses a few little problems:
* Adds a new option to `generate_CBED` that's necessary for two-beam simulations
* Corrects a bug where the Lobato potentials were not scaled correctly in the Bloch wave code
* Handles a bug where versions of h5py > 3 throw an error when reading empty PointList data
* Various performance and stability tweaks to batched CUDA disk detection


py4DSTEM/process/diskdetection/diskdetection_cuda.py
```diff
@@ -543,8 +542,8 @@ def get_maxima_2D(
                 Iy1 = ar[int(maxima["x"][i]), int(maxima["y"][i]) + 1]
                 deltax = (Ix1 - Ix1_) / (4 * Ix0 - 2 * Ix1 - 2 * Ix1_)
                 deltay = (Iy1 - Iy1_) / (4 * Iy0 - 2 * Iy1 - 2 * Iy1_)
-                maxima["x"][i] += deltax
-                maxima["y"][i] += deltay
+                maxima["x"][i] += deltax if np.abs(deltax) <= 1. else 0.
+                maxima["y"][i] += deltay if np.abs(deltay) <= 1. else 0.
```
bsavitzky, line 546: what's the new code here for?  why would the np.abs of deltax or deltay be larger than 1?  and if there are cases where this happens, do we need to be handling it some other way?

sezelt: It seems that after switching to float32 for the GPU version, sometimes rounding errors would cause `deltax` or `deltay` to be very large (several thousand), because in its definition we are dividing by the difference between very close values. So for now I am just catching that situation and skipping the subpixel update.

bsavitzky: Interesting, gotcha.  It might make sense to skip the subpixel update entirely, rather than for just a single direction.  Not sure how common this issue is, but it might also make sense to throw a warning message.  I would suggest at least adding a comment in the code indicating that this is happening.

bsavitzky: alrighty, merged.  This has also been included in v0.12.19, which is now current in PyPI

PR #272, opened 2022-05-26
TitanX auto read shape and CUDA fix

sezelt:
This PR does two things:
* Add the ability to read the metadata from TitanX datasets and auto-reshape the 3D data to 4D. This corrects a bug where bin-on-load was not possible for TitanX data because the memmap would be 3D.
* Fixes an index computation bug in the CUDA braggdisk detection that would cause illegal memory accesses on non-square data.


sezelt:
I am going to split this into different parts so I can add more to the CUDA changes and we can review them separately

PR #273, opened 2022-05-29
read TitanX tags inside DM files and auto-reshape 3D into 4D

sezelt:
The NCEM TitanX, for historical reasons, saves its 4D datasets as a 3D array. Luckily, it writes a custom set of metadata tags into the DM file that indicate the proper 4D shape of the data. This update checks for the existence of those tags and reshapes the data accordingly if they are present. This fixes a bug where bin-on-load failed for Titan data because of the shape mismatch. 

I don't have a non-square dataset easily on hand, so I'd appreciate if someone has one on hand so we can test that there is not an x/y flip in the conventions.

### DIFF
--- a/py4DSTEM/io/nonnative/read_dm.py
+++ b/py4DSTEM/io/nonnative/read_dm.py
@@ -51,6 +51,7 @@ def read_dm(fp, mem="RAM", binfactor=1, metadata=False, **kwargs):
                     dataSet = dmFile.getDataset(i)
                 i += 1
             dc = DataCube(data=dataSet["data"])
+            _process_NCEM_TitanX_Tags(dmFile, dc)
     elif (mem, binfactor) == ("MEMMAP", 1):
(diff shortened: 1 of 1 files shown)


bsavitzky (approved):

bsavitzky:
Hey Steve - I checked with a non-square image and it seemed to work just fine.

One thing I might add is a fix to the issue where the TitanX always rolls the data by two pixels, which could just be bundled with this code.  I can add that in, but wanted to check with you first - I would just add to your new function that's reading the tags a line at the end that performs the necessary rolling.  Wanted to check because this is an issue on essentially every TitanX dataset I see, but if it isn't present on some subset of TitanX data I'm unaware of we probably would not want to include this.

bsavitzky:
And in fact it does look like it is flipped for non-square data, glad I checked a second file.  Making that fix now

PR #274, opened 2022-05-29
CUDA-accelerated multicorr

sezelt:
This substantially reworks how multicorr is implemented in the CUDA-accelerated Bragg disk detection routine. The matrix-multiply DFT arrays are now generated by custom CUDA kernels, and the process is vectorized across all of the disk positions found in a single diffraction pattern. This gives an approximately 4x speedup over the previous implementation when using `unsample_factor=16`, and is now about half as fast as `upsample='poly'` at the same upsample factor. This PR also includes some bugfixes to the batched disk detection, including fixing an indexing error in the maximal points routine that was causing out-of-bounds memory access on the device for non-square diffraction patterns.

### DIFF
--- a/py4DSTEM/process/diskdetection/diskdetection_cuda.py
+++ b/py4DSTEM/process/diskdetection/diskdetection_cuda.py
@@ -585,12 +585,10 @@ def upsampled_correlation(imageCorr, upsampleFactor, xyShift):
             Upsampling factor. Must be greater than 2. (To do upsampling
             with factor 2, use upsampleFFT, which is faster.)
         xyShift:
-            Location in original image coordinates around which to upsample the
-            FT. This should be given to exactly half-pixel precision to
-            replicate the initial FFT step that this implementation skips
+            Array of points around which to upsample, with shape [N-points, 2]
 
     Returns:
-        (2-element np array): Refined location of the peak in image coordinates.
+        (N_points, 2) cupy ndarray: Refined locations of the peaks in image coordinates.
     """
 
     xyShift = (cp.round(xyShift * upsampleFactor) / upsampleFactor).astype(cp.float32)
@@ -603,7 +601,7 @@ def upsampled_correlation(imageCorr, upsampleFactor, xyShift):
 
     xSubShift, ySubShift = np.unravel_index(imageCorrUpsample.reshape(imageCorrUpsample.shape[0],-1).argmax(axis=1), imageCorrUpsample.shape[1:3])
 
-    # add a subpixel shift via parabolic fitting
+    # add a subpixel shift via parabolic fitting, serially for each peak 
     for idx in range(xSubShift.shape[0]):
         try:
             icc = np.real(
@@ -646,13 +644,13 @@ def dftUpsample(imageCorr, upsampleFactor, xyShift):
             Correlation image between two images in Fourier space.
         upsampleFactor (int):
             Scalar integer of how much to upsample.
-        xyShift (list of 2 floats):
+        xyShift (N_points,2) cp.ndarray, locations to upsample around:
             Coordinates in the UPSAMPLED GRID around which to upsample.
             These must be single-pixel IN THE UPSAMPLED GRID
 
     Returns:
         (ndarray):
-            Upsampled image from region around correlation peak.
+            Stack of upsampled images from region around correlation peak.
     """
     N_pts = xyShift.shape[0]
     imageSize = imageCorr.shape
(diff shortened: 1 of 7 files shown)


sezelt:
Btw for these new CUDA kernels, I implemented them in separate `.cu` files (mainly so that my text editor would do the right syntax highlighting). I think I made the necessary changes to `setup.py` so that the files get included in the PyPI distribution, but I'm not sure how to check that without going through the whole process of making a distribution.

ercius:
> I'm not sure how to check that without going through the whole process of making a distribution.

You could try using the pypi test repo:
https://packaging.python.org/en/latest/guides/using-testpypi/

Or try an alpha release by tagging with a version number similar to `1.0.0a1`

sezelt:
@ercius thanks for the tip! I tested it and the correct files are indeed getting packaged. 

@bsavitzky is there anything else we need to go over before this can be merged?

bsavitzky:
Hey Steve - I just need to take a quick pass through the code. I'll be able
to get to this by tomorrow!

py4DSTEM/process/diskdetection/multicorr_row_kernel.cu
```diff
@@ -0,0 +1,60 @@
+#include <cupy/complex.cuh>
+#define PI 3.14159265359
+extern "C" __global__
+void multicorr_row_kernel(
+	complex<float> *ar,
+	const float *xyShifts,
+	const long long N_pts,
+	const long long image_size_x,
+	const long long image_size_y,
+	const long long upsample_factor) {
```
alex-rakowski, line 10: Do these need to be `long long`?

py4DSTEM/process/diskdetection/diskdetection_cuda.py
```diff
@@ -671,39 +644,38 @@ def dftUpsample(imageCorr, upsampleFactor, xyShift):
...
+    blocks = ((np.prod(rowKern.shape) // multicorr_row_kernel.max_threads_per_block + 1),)
+    threads = (multicorr_row_kernel.max_threads_per_block,)
+    multicorr_row_kernel(blocks,threads,(rowKern, xyShift, N_pts, *imageSize, upsampleFactor))
+
+    # Apply the DFT arrays to the correlation image
     imageUpsample = cp.real(rowKern @ imageCorr @ colKern)
     return imageUpsample
 
-
+@numba.jit(nopython=True)
```
alex-rakowski, line 678: I think this can be written as a `numba.stencil` function, which I believe is even faster. https://numba.pydata.org/numba-doc/latest/user/stencil.html. 

Could also consider setting `target='gpu'` ?

py4DSTEM/process/diskdetection/diskdetection_cuda.py
```diff
@@ -120,7 +121,7 @@ def find_Bragg_disks_CUDA(
     if get_maximal_points.max_threads_per_block < DP.shape[1]:
         # naive blocks/threads will not work, figure out an OK distribution
         blocks = ((np.prod(DP.shape) // get_maximal_points.max_threads_per_block + 1),)
-        threads = get_maximal_points.max_threads_per_block
+        threads = (get_maximal_points.max_threads_per_block,)
```
alex-rakowski, line 124: curious why the tuple instead of an int?

py4DSTEM/process/diskdetection/diskdetection_cuda.py
```diff
@@ -326,13 +327,6 @@ def _find_Bragg_disks_single_DP_FK_CUDA(
     Returns:
         peaks                (PointList) the Bragg peak positions and correlation intensities
     """
-    assert subpixel in [
-        "none",
-        "poly",
-        "multicorr",
-    ], "Unrecognized subpixel option {}, subpixel must be 'none', 'poly', or 'multicorr'".format(
-        subpixel
-    )
```
alex-rakowski, line 335: Is this just a general purge of assert statements from py4DSTEM?

py4DSTEM/process/diskdetection/multicorr_row_kernel.cu
sezelt: Yes, the `int` type in Python maps to `long long` in CUDA

py4DSTEM/process/diskdetection/diskdetection_cuda.py
sezelt: I was not aware of stencil, that may come in handy in the future. However this function isn't called on the entire array so it doesn't make sense here. 

I'm not sure if it would be faster to come up with a CUDA solution for this particular interpolation operation or if launching a kernel for such a small unit of work would end up being slower than this rather quick CPU function, so I'm going to leave that for the future.

py4DSTEM/process/diskdetection/diskdetection_cuda.py
sezelt: `cupy` requires this to be passed as a tuple when invoking a kernel. It is possible to set up higher dimension grids of threads, in which case this tuple has 2 or 3 items, but for launches that require multiple thread blocks it ends up being cleaner to do this using 1D thread groups.

py4DSTEM/process/diskdetection/diskdetection_cuda.py
sezelt: This only affects the CUDA disk detection, where this was getting run for every single diffraction pattern, which is a waste of time. To be safe I'll add this assert into the top level `find_Bragg_disks` function so that it always definitely gets checked once.

sezelt:
I've pulled in the latest changes from dev and also added a little extra option that helps to free VRAM after the disk detection is done

bsavitzky (approved):

PR #278, opened 2022-06-09
Materials_project_api_update

alex-rakowski:


sezelt:
Why not just add it under the `acom` install option?

alex-rakowski:
That's a good idea, I had a couple of problems with versioning when I had it as a requirement but I think it was probably just a weird environment from trying different Pymatgen versions

sezelt:
If there are incompatibilities (which it sounds like is the case based on the Slack discussion) then I think you should specify versions for `pymatgen` and `mp_api` (or at least force the 2022 version of pymatgen and let pip do the rest).

alex-rakowski:
Added as specific versions, but we should try and losen them in the future. ... But this works as a patch:


```
import py4DSTEM
from mp_api import MPRester

with MPRester(key) as m:
  test_crys = m.get_structure_by_material_id("mp-48")
  
xtal = py4DSTEM.process.diffraction.Crystal.from_pymatgen_structure(test)

xtal.calculate_structure_factors(k_max=4.0, tol_structure_factor=0.0

xtal = py4DSTEM.process.diffraction.Crystal.from_pymatgen_structure('mp-48')

xtal.calculate_structure_factors(k_max=4.0, tol_structure_factor=0.0)

....

```

sezelt:
Is there a reason for forcing these specific versions? We do not want to go back to having hardlocked dependencies unless there is an ironclad reason for it, given how much trouble it caused in the past.

alex-rakowski:
They are the ones that I know work and restore the functionality.  I couldn't get non specific versions to install correctly in a fresh envrionment (maybe M1 related?) I could change it to `'pymatgen >=2022.5.26'` and `'mp-api >= 0.24.1'` which could help with future releases. But I'd need to do more testing to loosen them up beyond this.

alex-rakowski:
I've done some more testing and the loosest I could get to work is 'pymatgen >=2022' and 'mp-api == 0.24.1', which works for python 3.8 and 3.9.  I don't think pymatgen supports python3.7.  I can do some more checks in the future but this currently works and restores functionality.

sezelt:
I did some investigating and I think the problems you were running into earlier may have been caused by using an outdated MP API key. The new interface has a new format for the keys, and the website generates new ones by default. By going to the legacy MP page and getting an API key from there, the old MPRester interface we are using still works fine.

What problems are you facing when using looser dependency specs? Is the interface not working, or does it not resolve?

alex-rakowski:
> I did some investigating and I think the problems you were running into earlier may have been caused by using an outdated MP API key. The new interface has a new format for the keys, and the website generates new ones by default. By going to the legacy MP page and getting an API key from there, the old MPRester interface we are using still works fine.


I'm using my MP API Key from the dashboard on the new materials project website. I can't seem to log in to the legacy website with email or Github. I'm not sure how much value there is in sticking with the old API either 


I've pasted my simplest reproducible error 


```
from pymatgen.ext.matproj import MPRester
key = <string>  # from materials project website 
m = MPRester(key)
m.get_structure_from_material_id('mp-48')
--> 

MPRestError: REST query returned with error status code 403. Content: b'{"valid_response": false, "error": "API_KEY is not a valid key.", "version": {"db": "2020_09_08", "pymatgen": "2022.0.8", "rest": "2.0"}, "created_at": "2022-06-10T13:32:16.109451"}'


from mp_api import MPRester
key = <string>  # same key as above
m = MPRester(key)
m.get_structure_by_material_id("mp-48")

-->
Structure Summary
Lattice
    abc : 2.4677242827325747 2.4677242827325747 8.685038
 angles : 90.0 90.0 120.00000758003493
 volume : 45.80317575176706
      A : 1.233862 -2.137112 0.0
      B : 1.233862 2.137112 0.0
      C : 0.0 0.0 8.685038
PeriodicSite: C (0.0000, 0.0000, 6.5138) [0.0000, 0.0000, 0.7500]
PeriodicSite: C (0.0000, 0.0000, 2.1713) [0.0000, 0.0000, 0.2500]
PeriodicSite: C (1.2339, 0.7124, 6.5138) [0.3333, 0.6667, 0.7500]
PeriodicSite: C (1.2339, -0.7124, 2.1713) [0.6667, 0.3333, 0.2500]

```



> 
> What problems are you facing when using looser dependency specs? Is the interface not working, or does it not resolve?

When loosening the dependency on `mp-api` neither `>=` or `<=` resolve. I'm not sure why I think one of the first issues is with `Kiwisolver`? 

I think this worked, but `mp-api==0.24.2` was released 13 hours ago and I can't remember if I've checked it since. 

```
conda create ...
...
pip install mp-api
pip install py4dstem[acom]
```

bsavitzky (approved):

PR #286, opened 2022-06-23
h5py minimum requirement update

jordiferrero:
As discussed in Slack, yesterday I ran into the problem of trying to read a `.h5` file containing a `braggpeaks_centered` `PointListArray` object stored with other files.
After discussion, I found out the problem was caused by the h5py versions 3.0 and 3.1.
Hence I recommend updating the minimum requirement to v.3.2.0 at least.

- [x] Ready for review


sezelt (approved):
Looks good to me. I have been using 3.6.0, which seems to be the latest, for a while and have not had any compatibility problems

PR #288, opened 2022-06-28
Return S Matrix from Bloch wave diffraction

sezelt:
This adds a new option, `return_Smatrix`, to `generate_dynamical_diffraction_pattern`, which returns the scattering matrix and the corresponding incident plane wave vector instead of the diffraction intensities.

### DIFF
--- a/py4DSTEM/process/diffraction/crystal_bloch.py
+++ b/py4DSTEM/process/diffraction/crystal_bloch.py
@@ -372,16 +372,6 @@ def generate_dynamical_diffraction_pattern(
         g.T, foil_normal=-foil_normal @ zone_axis_rotation_matrix
     )
 
-    # import matplotlib.pyplot as plt
-    # sgp = np.sign(sg) >= 0
-    # c = np.zeros_like(g)
-    # c[sgp,:] = np.array([1,0,0])
-    # c[~sgp,:] = np.array([0,0,1])
-    # fig,ax = plt.subplots(dpi=200)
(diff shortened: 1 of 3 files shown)


py4DSTEM/process/diffraction/crystal_bloch.py
```diff
@@ -406,22 +418,26 @@ def generate_dynamical_diffraction_pattern(
...
     # I = |psi|^2 ; psi = C @ E(z) @ C^-1 @ psi_0, where E(z) is the thickness matrix
-    if return_complex: 
-  
-        # calculate the amplitudes 
+    if return_Smatrix:
+        Smatrices = [
+            C @ np.diag(np.exp(2.0j * np.pi * z * gamma)) @ C_inv
+            for z in np.atleast_1d(thickness)
+        ]
+        return Smatrices, psi_0
```
alex-rakowski, line 426: Is it worth having the ability to generate the `pointlist` object as usual and then return  ~ `pointlist_obj, Smatricies_list, psi_0`, or spin this into a new function `generate_dynamical_Smatricies`  ?

sezelt: Someday in the future I do want to make the Bloch code more modular, in which case this will become a separate function, but I don't want to tackle that problem yet. For now, I think this way is clean enough.

alex-rakowski (approved):
Looks good. In the future, it would be good to spin this out into a separate function, as syntax-wise it's currently a bit muddled.

PR #292, opened 2022-07-12
Fixes for Bloch notebook

sezelt:
Import fixes and add a new `PointList` data convenience function

### DIFF
--- a/py4DSTEM/io/__init__.py
+++ b/py4DSTEM/io/__init__.py
@@ -1,5 +1,5 @@
 from . import datastructure
-from .datastructure import DataCube, PointListArray, PointList
+from .datastructure import *
 from .read import *
 from .native.read import print_h5_tree
 from .native.write import save
--- a/py4DSTEM/io/__init__.py
+++ b/py4DSTEM/io/__init__.py
@@ -1,4 +1,5 @@
 from . import datastructure
+from .datastructure import DataCube, PointListArray, PointList
 from .read import *
 from .native.read import print_h5_tree
 from .native.write import save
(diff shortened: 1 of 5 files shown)


py4DSTEM/io/__init__.py
```diff
@@ -1,4 +1,5 @@
 from . import datastructure
+from .datastructure import DataCube, PointListArray, PointList
```
cophus, line 2: I changed it to:

from .datastructure import *

Is there any risk in importing all of them?

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -345,9 +361,7 @@ def from_unitcell_parameters(
 
         return Crystal.from_pymatgen_structure(structure)
 
-    def setup_diffraction(
-        self, accelerating_voltage: float
-    ):
+    def setup_diffraction(self, accelerating_voltage: float):
```
cophus, line 364: I still think we should be able to skip specifying voltage here in many quick workflows - perhaps instead specify a default option for crystal on instantiation?

cophus (commented):
Looks good - just check the specific data structures vs *

py4DSTEM/io/__init__.py
sezelt: I think it is fine

py4DSTEM/process/diffraction/crystal.py
sezelt: OK I made it default to 300 kV if none is set, and give a warning in `generate_diffraction_pattern`

py4DSTEM/io/__init__.py
sezelt: Made the change and pushed it

cophus (approved):
Looks good!

bsavitzky (approved):

PR #294, opened 2022-07-12
Fix multicorr

sezelt:
The wrong array was getting passed around when multicorr was selected

### DIFF
--- a/py4DSTEM/process/calibration/braggvectors.py
+++ b/py4DSTEM/process/calibration/braggvectors.py
@@ -3,7 +3,6 @@ import numpy as np
 from typing import Optional, Union
 from numbers import Number
 
-from ...io.datastructure import BraggVectors
 from ...io.datastructure.emd import PointListArray
 from ...io.datastructure.py4dstem import Calibration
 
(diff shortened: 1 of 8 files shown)


py4DSTEM/process/calibration/braggvectors.py
```diff
@@ -100,6 +101,10 @@ def center_braggpeaks(
     Returns:
         (PointListArray): the centered Bragg peaks
     """
+
+    if isinstance(braggpeaks, BraggVectors):
+        braggpeaks = braggpeaks.vectors_uncal
+
```
bsavitzky, line 107: This has to go.  These functions change the underlying data in the PointListArray, and braggpeaks.vectors_uncal should *never* be modified.  It needs to be copied first, and the output passed in to this function, as when you run braggpeaks.calibrate().

bsavitzky (changes requested):
One change requested - see comments

py4DSTEM/process/calibration/braggvectors.py
sezelt: OK I think I misunderstood how this gets called, since I was adapting an old notebook. So just to be sure, this function should never actually get a `BraggVectors` object as an argument? And thus I should just drop these three lines fully?

bsavitzky: Yup, you got it.  Really this function should only be called internally through `BraggVectors.calibrate()`, which will handle first copying the underlying uncalibrated PLA, modifying the new, copied PLA with these functions, and then replacing the calibrated vectors attribute with this new PLA.

PR #295, opened 2022-07-12
More imports to make the tutorials work

cophus:
Very minor changes


sezelt (approved):
I think this is fine... It seems to not have `diffraction` imported but I think it will get pulled in from dev once it's merged

sezelt:
OK confirmed, `diffraction` is still imported after merging

PR #297, opened 2022-07-13
Legacy I/O in v13

sezelt:
This _should_ make the majority of legacy EMD files readable. I haven't tested every combo of version and object type, but the handful of files I've tested are working


bsavitzky (approved):
Yes!!  Thanks Steve.  This is greeeaaaaaaaat

PR #298, opened 2022-07-14
fix multicorr bug

smribet:


sezelt:
In my initial multicorr fix, I did not realize the other situations where `get_maxima_2D` was getting called and chose a clumsy way of specifying the arguments. This changes where the Fourier transforms are happening to make it behave as expected

sezelt (approved):

PR #299, opened 2022-07-14
adding .show to top level namespace

alex-rakowski:


alex-rakowski:
This allows `py4DSTEM.show(...)` and `py4DSTEM.visualize.show(...)` as ways to access `show` function

PR #303, opened 2022-07-15
07-08 acom notebook

alex-rakowski:
Changes required to run 07-08 notebooks.


sezelt:
I don't understand these changes at all. You've added 8 new files that are (modified?) copies of old ones but with "2" added to the name, so the diffs are meaningless, and they aren't even imported anywhere.

alex-rakowski:
Not sure why `* 2.py` were created but I've removed them now, but the PR looks correct now.

py4DSTEM/preprocess/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .preprocess import *
 from .darkreference import *
-from .electroncount import *
+# from .electroncount import * # get_maximal_points not in utils anymore will break
```
sezelt, line 3: This broken import is fixed in #304

py4DSTEM/preprocess/relativity.py
```diff
@@ -16,8 +16,8 @@
 import matplotlib.pyplot as plt
 from ipywidgets import FloatProgress
 from time import time
-from ..utils import get_shift
-from ...io import DataCube
+from ..process.utils import get_shift
+from ..io.datastructure.py4dstem import DataCube
```
sezelt, line 20: I think you should always import `py4DSTEM.io.DataCube`. We've had problems before where multiple instances of the class, with different `id`, get created due to importing through different paths. When this happens, `assert isinstance(datacube, py4DSTEM.io.DataCube` can fail

py4DSTEM/preprocess/electroncount.py
```diff
@@ -7,7 +7,8 @@
 import numpy as np
 from scipy import optimize
 
-from ..utils import get_maximal_points, print_progress_bar, bin2D
+# get_maximal_points has gone missing / renamed will break
```
sezelt, line 10: This broken import is fixed by #304

sezelt (changes requested):
I think the only necessary change is the `..io.datastructure.py4dstem.DataCube` import thing.

PR #304 fully corrects some of the `preprocess` bugs that this hacks, which will cause merge conflicts. So I think we should merge #304 first and then resolve the conflicts on this one.

bsavitzky (commented):
This looks good to me!!  Following Steve's comment, I'll wait until we can discuss and finish PR #304, and then can come back and finish pulling this code in

bsavitzky (approved):

PR #304, opened 2022-07-16
re-enable some nonnative file capabilities

sezelt:
This brings back most (but not all) of the previous nonnative file handling. 
* DM reading, both RAM and MEMMAP, but not bin-on-load, is tested working
* EMPAD reading is re-linked but untested (test files needed)
* K2 reading is re-linked but untested (will test later)
* Relativity-MRC reading is fully removed
* Kitware-H5 reading is disabled since `CountedDataCube` is still missing. My preference is to remove this since I do not think it is compatible with the latest versions of their files anyway. 

Nonnative files are now read using a separate function than native files, `py4DSTEM.io.import_`

The DM reader now supports 2D and 4D data transparently, and returns a `DataCube` for 4D data and an `Array` for 2D data.


py4DSTEM/process/utils/utils.py
```diff
@@ -260,31 +260,6 @@ def add_to_2D_array_from_floats(ar, x, y, I):
     ar[x1[mask], y1[mask]] += (    dx[mask]) * (    dy[mask]) * I[mask]
     return ar
 
-
-def print_progress_bar(iteration, total, prefix='', suffix='', decimals=1,
```
alex-rakowski, line 264: Has this func been removed/replaced with Tqdmnd?

py4DSTEM/io/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    if fext in [
+        ".h5",
+        ".H5",
+        ".hdf5",
+        ".HDF5",
+        ".py4dstem",
+        ".py4DSTEM",
+        ".PY4DSTEM",
+        ".emd",
+        ".EMD",
```
alex-rakowski, line 16: I don't think the accepted file extensions are described in any of the doc strings,  maybe raise in the exception what the expected types are?

py4DSTEM/io/utils.py
```diff
@@ -0,0 +1,37 @@
+from os.path import splitext
+
+def parse_filetype(fp):
```
alex-rakowski, line 3: might be worth doing string santizing e.g `fext.lower()` incase someone does `* .dM3`

py4DSTEM/preprocess/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .preprocess import *
 from .darkreference import *
 from .electroncount import *
-from .relativity import *
```
alex-rakowski, line 4: has this functionality been removed or moved elswhere?

py4DSTEM/preprocess/electroncount.py
```diff
@@ -74,15 +74,13 @@ def electron_count(datacube, darkreference, Nsamples=40,
...
-                print_progress_bar(Rx*R_Ny+Ry+1, R_Ny*R_Nx, prefix=' Counting:', suffix='Complete',
-                                                                                 length = 50)
                 frame = datacube[Rx,Ry,:,:].astype(np.int16)    # Get frame from file
                 workingarray = frame-darkreference              # Subtract dark ref from frame
                 events = workingarray>thresh_bkgrnd             # Threshold electron events
                 events *= thresh_xray>workingarray
 
                 ## Keep events which are greater than all NN pixels ##
-                events = get_maximal_points(workingarray*events)
+                events = get_maxima_2D(workingarray*events)
```
alex-rakowski, line 83: is this the same functionality, if so we should replace its use elsewhere

py4DSTEM/process/dpc/dpc.py
```diff
@@ -209,7 +209,7 @@ def get_rotation_and_flip_maxcontrast(CoMx, CoMy, N_thetas, paddingfactor=2,
                                           n_iter=n_iter)
         stds[i] = np.std(phase)
         if verbose:
```
alex-rakowski, line 211: if verbose = True results in pass, should it be stripped out, or is the plan to restore a progress bar?

py4DSTEM/preprocess/electroncount.py
```diff
@@ -98,15 +96,13 @@ def electron_count(datacube, darkreference, Nsamples=40,
...
-                print_progress_bar(Rx*R_Ny+Ry+1, R_Ny*R_Nx, prefix=' Counting:',
-                                    suffix='Complete', length=50)
                 frame = datacube[Rx,Ry,:,:].astype(np.int16)    # Get frame from file
                 workingarray = frame-darkreference        # Subtract dark ref from frame
                 events = workingarray>thresh_bkgrnd       # Threshold electron events
                 events *= thresh_xray>workingarray
 
                 ## Keep events which are greater than all NN pixels ##
-                events = get_maximal_points(workingarray*events)
+                events = get_maxima_2D(workingarray*events)
```
alex-rakowski, line 105: Is this the same functionality, if so we should replace its use elsewhere?

py4DSTEM/io/nonnative/read_dm.py
```diff
@@ -4,16 +4,15 @@
 from pathlib import Path
 from ncempy.io import dm
 from ..datastructure import DataCube, Array, Metadata
-#from ...preprocess.utils import bin2D
+from ...process.utils import bin2D
```
alex-rakowski, line 9: This looks correct and doesn't raise any import errors but I haven't checked the against un/accepted file types

py4DSTEM/io/read.py
```diff
@@ -6,19 +6,15 @@
 import h5py
```
alex-rakowski, line 7: This looks correct and doesn't raise any import errors but I haven't checked the against un/accepted file types

alex-rakowski (commented):
Some minor suggestions, but it looks correct as far as I can tell.

py4DSTEM/process/utils/utils.py
sezelt: yes

py4DSTEM/io/utils.py
sezelt: updated

py4DSTEM/io/utils.py
sezelt: ok

py4DSTEM/preprocess/__init__.py
sezelt: removed, as noted in the description of the PR

py4DSTEM/preprocess/electroncount.py
sezelt: The function name just changed

py4DSTEM/process/dpc/dpc.py
sezelt: I replaced it with tqdm

py4DSTEM/preprocess/electroncount.py
sezelt: just a name change. The old name isn't being used anywhere else

sezelt:
Review comments are addressed, can someone take a look/approve?

bsavitzky (commented):
This all looks good to me!  Except....I don't love splitting up the `read` function into two functions for native/non-native files.  Was this done for simplicity of internal structure, or for some other reason?  What do folks think about having this functionality moved back into `read`, such that the filetype is parsed and the appropriate reading function is called?

bsavitzky (approved):

PR #306, opened 2022-07-18
get_probe_size and read file fixes

smribet:


sezelt (approved):
I'm gonna fast track this since it fixes the reader bug but making some of the `assert` error messages more verbose would be good to have in a future update

PR #307, opened 2022-07-19
Cuda fix

alex-rakowski:


cophus (commented):
Looks good to me!  Someone else should probably double check though.

cophus (approved):

PR #308, opened 2022-07-19
Improving probe kernel plotting

cophus:
Plot all 4 quadrants of the probe, 2 line traces


bsavitzky (approved):
made two fixes:
- made sure BraggVectors._v_uncal had its .name set to "_v_uncal"
- added an optional `name` argument to BraggVectors.copy()

PR #309, opened 2022-07-19
fix for saving plus a little tweak in Bloch

sezelt:
This makes sure that when a `BraggVectors` object gets saved, the `PointListArray`s inside have the correct names, so that the reader can find them later.

Because I happened to be working on it when the saving bug arose, this also includes the ability to specify different thermal displacements per-element in the Bloch code


bsavitzky (approved):

bsavitzky:
minor fixes + solved merge conflict

PR #310, opened 2022-07-19
Config checker

alex-rakowski:
Still needs a few more changes but feedback would be useful.

Changes required before merge:
- move `configuration_checker.py` and `tqmnd.py` into new directory
- change `__init__.py`
- change modules list in `configuration_checker.py`to things more accurately reflect required/expected modules


sezelt:
This is really cool! I think it would be good to group the checks under header based on what features require what. So for example, have a ACOM section that checks pymatgen, a ML section that has tensorjax or whatever witchcraft, an acceleration section that looks for numba and cupy, etc. 
Also checking for numpy is probably redundant since the module won't import at all without it

alex-rakowski:
I've added the module checking feature, and added top level utils for `tqmnd.py` and `configuration_checker.py`. Still need change the verbose behaviour and the ordering of the outputs.

bsavitzky (approved):
I removed some old comment/TODO's, otherwise looks great

PR #311, opened 2022-07-20
Calibration fixes

cophus:
Some quick and dirty methods to allow masking for the origin fits.


bsavitzky (approved):
noooooooooooooo

ok, as a temporary fix i'm merging this, but it makes me sad

PR #317, opened 2022-07-25
changes to Bloch defaults

sezelt:
Change the default parameters in Bloch wave to make sure that forbidden beams are included by default


sezelt:
Converting this to a draft since we have decided it's a post-M&M item

sezelt:
Closing as these changes are also part of #318

PR #318, opened 2022-07-28
move MPRester import

sezelt:
Move the MPRester import so that it doesn't get called when passing a `Structure` object


sezelt:
I decided to use this PR as a test for the new auto-linter, so there are some edits related to that in here now too. With these changes it is working as intended: It only checks the handful of errors that I specified in the config file, and it only does it on the files that have been edited by the current PR

sezelt:
This now also fixes #326

bsavitzky (approved):

PR #320, opened 2022-08-09
Ring pattern

smribet:


py4DSTEM/process/diffraction/crystal.py
```diff
@@ -621,7 +622,112 @@ def generate_diffraction_pattern(
...
+                )
+
+            intensity.append(pattern['intensity'])
+            radii.append((pattern['qx']**2 +  pattern['qy']**2)**0.5)
+        
+        intensity = np.concatenate(intensity)
+        radii = np.concatenate(radii)
 
+        radii_unique, index, counts, = np.unique(radii, return_counts = True, return_index = True)
+        intensity_unique = intensity[index] * counts
```
sezelt, line 720: I think that this is not summing the intensities in the right way. The `index` returned by `unique` only gives the _first_ occurrence of the value, not all of them. I think you need to use the `return_inverse` option to `unique` and then use `bincount` for the accumulation. The code below should show the problem, but I think that to make it work correctly in our case we will need to specify the `minlength` option for `bincount`...

```
In [1]: import numpy as np

In [2]: r = np.array([1,2,2,3,4,4,5])

In [3]: i = np.array([1,1,1,2,2,3,3])

In [4]: r
Out[4]: array([1, 2, 2, 3, 4, 4, 5])

In [5]: i
Out[5]: array([1, 1, 1, 2, 2, 3, 3])

In [6]: rad,idx,cts = np.unique(r, return_counts=True, return_index=True)

In [7]: i[idx]*cts
Out[7]: array([1, 2, 2, 4, 3])

In [8]: rad
Out[8]: array([1, 2, 3, 4, 5])

In [9]: idx
Out[9]: array([0, 1, 3, 4, 6])

In [10]: rad,idx,inv,cts = np.unique(r, return_counts=True, return_index=True,return_inverse=True)

In [11]: inv
Out[11]: array([0, 1, 1, 2, 3, 3, 4])

In [12]: i[inv]
Out[12]: array([1, 1, 1, 1, 2, 2, 2])

In [13]: np.bincount(inv,weights=i)
Out[13]: array([1., 2., 2., 5., 3.])
```

sezelt (changes requested):
Looks good apart from the potential mistake summing the intensities

py4DSTEM/process/diffraction/crystal.py
smribet: Thanks for finding this error @sezelt!

sezelt (approved):

PR #323, opened 2022-08-16
Add pylint action file

sezelt:
Add a github action based on [this](https://github.com/py-actions/flake8) that runs `flake8` with a subset of fatal errors selected to warn about when people PR into dev


sezelt:
This seems to now work to do a check for major errors whenever someone makes a pull request, only checking files that are new or edited. One thing I am not sure about is if I put the configuration file for flake8 in the right place. If that turns out to be wrong, it'll just report a lot of stylistic errors and we can figure out where to move it later.

bsavitzky (approved):

PR #327, opened 2022-08-23
Virtual detectors update and complex plotting 

smribet:
This PR is to help simply `virtualimage.py` and associated `datacube_fns` and `io` files. There are also functionalities for shifting the center of the detector and turning off the progress bar. Lastly, there's a new `vis_special` function for plotting complex arrays (example below).  Look forward to discussing more!  

~Stephanie 

```
AA = 200

test_object = np.ones((AA,AA), dtype = 'complex')

kx = np.fft.fftshift(np.fft.fftfreq(AA))
ky = np.fft.fftshift(np.fft.fftfreq(AA))
kya,kxa = np.meshgrid(ky,kx)


test_object *= np.exp(1j*np.sin(kxa*4*np.pi))
test_object *= np.cos(kya * np.pi*5)

fig, ax = py4DSTEM.visualize.show_complex(
    test_object*10, 
    returnfig = True, 
    cbar = True, 
    circle = {
      'center':(60, 60),
      'R': 15,
      'fill':True
    }
)

```

test_object_complex


bsavitzky (commented):
Hey Stephanie - this looks great, thanks for these edits!  This is great progress.

I think this module could still use some work - perhaps you, me, and @alex-rakowski could all sit down at some point, as the main file in the module is now a bit of a palimpsest of all of our not-always-aligned efforts 😂 

Three things that jump out at me at this point:

(1) in making the masks, it looks like all of the masks now are computed over the full fields of view.  Alex R earlier implemented a faster approach e.g. for circular and annular detectors, present in the last version of virtualimage.py, which only performed the calculations on a small subset of the image under the circle's radius, then just plopped that into an existing boolean array of zeros.  For making a single mask at the beginning of an entire virtual image computation this shouldn't matter - however for e.g. a mask that shifts with the scan position, this will likely save quite a bit of time and I think it worth keeping, if we can finagle the architecture so that it, um, makes sense.

(2) it looks like you've removed the `_centered` and `_calibrated` mode options.  I think this makes sense - the modes you've included really are the only ones we should have for now I think - however, I do like the option these modes were meant to make user-accessible, think it would be good to include them and that we could do so by adding two flags (`centered = False`, `calibrated = False`).  These options are meant to refer to the reference/calibration of the `geometry` variable.  I.e. with both set to False, geometry refers to the raw detector pixel positions.  With `centered = True` geometry is in units of pixels and the center is set to the center of the detector, and with `calibrated=True` the geometry is in real units.  If we do implement this there are some fine points to discuss, maybe easiest to talk directly to first see if we can agree on the interface and then to discuss implementation details.

(3) it looks like you began this PR before the absolute imports update, so we'll have merge conflicts there - we'll want to go with the absolute imports, but that can be easily handled later on at merge time

sezelt:
It looks like the linter worked on this one, and found an actual bug!

smribet:
@bsavitzky thanks for all the feedback! I'd definitely like to meet with you and @alex-rakowski to figure out how to align all these efforts. I can fix the imports for now and will wait on the other two until we have a chance to discuss more. 

@sezelt That's awesome! I fixed the bug.

bsavitzky (commented):
just one comment here following today's dev team meeting

bsavitzky (approved):
I made one small change - now the `shift_center` default in the datacube method is True iff `centered` is set to True, rather than if both `centered` and `calibrated` are True.

PR #331, opened 2022-09-07
Fixes for ACOM and other misc functions

cophus:
Mostly fixing ACOM bugs


sezelt (approved):
Functionality seems good, I just made some tweaks to the docstrings and applied some formatting

PR #332, opened 2022-09-08
add an import test

sezelt:


py4DSTEM/process/diskdetection/diskdetection_aiml_cuda.py
```diff
@@ -219,7 +219,8 @@ def _find_Bragg_disks_aiml_single_DP_CUDA(DP, probe,
                                   peaks = None,
                                   get_maximal_points = None,
                                   blocks = None,
-                                  threads = None):
+                                  threads = None,
+                                  model_path="path/to/model"):
```
bsavitzky, line 223: not sure if "path/to/model" makes sense as a default?

bsavitzky (approved):

py4DSTEM/process/diskdetection/diskdetection_aiml_cuda.py
sezelt: @alex-rakowski  can you recommend a better solution here? `model_path` was being used but never defined, so I just added it to the argument list with a nonsense default as a stopgap to remove the syntax error

bsavitzky: I just changed it to `None` - this will have the correct behavior when passed to _get_latest_model from diskdetection_aiml.py

PR #333, opened 2022-09-08
Fix DM binning on load and read calibrations from metadata

sezelt:
This PR re-adds the bin-on-load feature for DM files and grabs the pixel calibrations using `ncempy`.
I've tested this with TitanX and K3 data and it seems to work as intended. 

One remaining issue is that the `__repr__` method for `Array` is not aware of these calibrations, so the printout of the `DataCube` still shows pixel units


sezelt:
Converting this to a draft since it is broken by some other DM files that we have

sezelt:
A sampling of miscellaneously corrupted DM files are now working, so I'm reopening this

py4DSTEM/io/datastructure/py4dstem/calibration.py
```diff
@@ -132,7 +132,7 @@ def set_R_pixel_size(self,x):
     def get_R_pixel_size(self):
         return self._get_value('R_pixel_size')
     def set_Q_pixel_units(self,x):
-        pix = ('pixels','A^-1')
+        pix = ('pixels','A^-1','mrad')
```
bsavitzky, line 135: not sure if `'mrad'` is our best option, maybe should consider `'arcmilliseconds'` instead

bsavitzky (approved):

PR #337, opened 2022-09-13
adds new io defaults

bsavitzky:
Edits the default behavior of the `read` function such that if `read(filepath)` is called with no additional parameters specified:

- if the file contains only one block of data, finds and returns it
- if the file contains multiple top level groups (e.g. '4DSTEM_experiment' and '4DSTEM_simulation'), raises a warning and returns those two group names
- if the file contains a single top level group (e.g. the default, '4DSTEM') but this group contains multiple objets in its own top level, raises a warning and returns a list of valid h5 paths to those objects (e.g. ['4DSTEM/datacube', '4DSTEM/virtual_image'])

All other behavior is unchanged, so `Tree=True` is still default, meaning if a single block of data which contains a large tree of data underneath it, all that data will be saved/loaded by default using `save(path,data)`/`read(path)`

### DIFF
--- a/py4DSTEM/io/google_drive_downloader.py
+++ b/py4DSTEM/io/google_drive_downloader.py
@@ -9,19 +9,33 @@ import os
 # single files
 sample_file_ids = {
     'FCU-Net' : '1-KX0saEYfhZ9IJAOwabH38PCVtfXidJi',
-    'sample_diffraction_pattern':'1ymYMnuDC0KV6dqduxe2O1qafgSd0jjnU'
+    'sample_diffraction_pattern':'1ymYMnuDC0KV6dqduxe2O1qafgSd0jjnU',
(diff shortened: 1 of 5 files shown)


smribet:
I'm not sure this is compatible with the new 0.13 tutorial notebooks. For notebook 03 when you read the dataset
`dataset = py4DSTEM.read(file_path_data)` the `read` function now returns `['4DSTEM_experiment\\data', '4DSTEM_experiment\\metadata']` instead of a datacube. Even with a suggested `root` selected, you still need the legacy `data_id` syntax to load the dataset.

bsavitzky:
Thanks @smribet - it looks like there was a bug when reading legacy files.  I just pushed an update that should fix this.

smribet:
The new `read` function looks good to me! I made some tiny suggested changes to the doc strings for the `kernel` functions.

smribet (approved):

PR #338, opened 2022-09-13
"Fix" for DM pixel size reading

sezelt:
@smribet noticed that the updates for reading pixel sizes from DM files breaks on 2D images. This patches the code so that we don't even attempt to read the pixel sizes for 2D images. I'm not certain how to robustly tell from the tags if a 2D image is real or reciprocal, and set thus how to use the pixel sizes appropriately, so I'm kind of kicking the can down the road here.

### DIFF
--- a/py4DSTEM/io/nonnative/read_dm.py
+++ b/py4DSTEM/io/nonnative/read_dm.py
@@ -42,6 +42,7 @@ def read_dm(filepath, name="dm_dataset", mem="RAM", binfactor=1, **kwargs):
                 break
 
         # We will only try to read pixel sizes for 4D data for now
+        pixel_size_found = False
         if dmFile.dataShape[dataset_index] > 2:
             # The pixel sizes of all datasets are chained together, so
             # we have to figure out the right offset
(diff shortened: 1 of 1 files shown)


sezelt:
@smribet can you test this on the dataset that failed earlier?

py4DSTEM/io/nonnative/read_dm.py
```diff
@@ -41,35 +41,37 @@ def read_dm(filepath, name="dm_dataset", mem="RAM", binfactor=1, **kwargs):
...
-        # Convert mrad to Å^-1 if possible
-        if Q_pixel_units == "mrad":
-            voltage = [v for t,v in dmFile.allTags.items() if "Microscope Info.Voltage" in t]
-            if len(voltage) == 1:
-                from py4DSTEM.process.utils import electron_wavelength_angstrom
-                lamda = electron_wavelength_angstrom(voltage[0])
-                Q_pixel_units = "A^-1"
-                Q_pixel_size = Q_pixel_size / lamda / 1000. # convert mrad to 1/Å
+        # We will only try to read pixel sizes for 4D data for now
+        if dmFile.dataShape[dataset_index] > 2:
```
bsavitzky, line 45: will this not break for 3D stacks? perhaps this should be `== 4`?

smribet:
> @smribet can you test this on the dataset that failed earlier?

It works!

py4DSTEM/io/nonnative/read_dm.py
sezelt: Titan data is 3-D inside the file, so we let it pass here. If the special TitanX tags are found it reshapes the data to 4D, and then at a later step we enforce `==4`

bsavitzky: ah gotcha, the `try...except` statements take care of it

bsavitzky (approved):

PR #340, opened 2022-09-17
Change how Orix file writer computes data

sezelt:
The Orix file writer in ACOM requires the use of Euler angles. It seems that there is some inconsistency between the Euler angles and the orientation matrices generated by ACOM, which I don't quite understand, but which are related to the inversion symmetry operation. This PR changes the Orix file writer to use `scipy` to convert the (presumably more correct) orientation matrices into Euler angles. 

I am not 100% sure of all the conventions that should be used in this conversion. Right now, I transpose the matrix, and in the case of the mirror flag being set also multiply it by an xy swap. This gives Euler angles that feel like they have good vibes, but we need to test it more rigorously first.


sezelt:
taking this out of draft now that the vibe checks on the orientations have passed

cophus:
I think this correction works, now that the ACOM module has been corrected to not produce improper rotations.

cophus (approved):

PR #342, opened 2022-09-19
automagic propagation of calibrations

sezelt:
Using some intense Python from [this StackOverflow](https://stackoverflow.com/questions/30104047/how-can-i-decorate-an-instance-method-with-a-decorator-class), this adds a magical decorator to several methods of the `Calibration` class that allows it to call the `calibrate()` methods of associated objects (_targets_, in our parlance). Objects whose `calibration` method should be called are registered with the `Calibration` instance using `register_target` (and unregistered with `unregister_target`. As of now, I have not added any of these links be default, pending some testing.

[This other StackOverflow](https://stackoverflow.com/questions/3798835/understanding-get-and-set-and-python-descriptors) is also potentially useful for understanding how this works.


sezelt:
Among the next steps for this PR is to add a `calibrate` method to the `DataCube` class such that the dim vectors get updated based on the calibrations (and perhaps to, by default, link it for automagic calibration)

py4DSTEM/io/datastructure/py4dstem/calibration.py
```diff
@@ -123,20 +128,25 @@ def get_Rshape(self,x):
...
     def get_Q_pixel_size(self):
         return self._get_value('Q_pixel_size')
+
+    @propagating_calibration
     def set_R_pixel_size(self,x):
         self._params['R_pixel_size'] = x
     def get_R_pixel_size(self):
         return self._get_value('R_pixel_size')
+
     def set_Q_pixel_units(self,x):
```
bsavitzky, line 143: i think the `@propagating_calibration` decorator be applied to setting the R/Q pixel units as well as sizes

py4DSTEM/io/datastructure/py4dstem/calibration.py
```diff
@@ -147,22 +157,29 @@ def set_qx0(self,x):
...
         self._params['qx0_meas'] = x
     def get_qx0_meas(self,rx=None,ry=None):
         return self._get_value('qx0_meas',rx,ry)
+
     def set_qy0_meas(self,x):
         self._params['qy0_meas'] = x
     def get_qy0_meas(self,rx=None,ry=None):
         return self._get_value('qy0_meas',rx,ry)
+
+    @propagating_calibration
```
bsavitzky, line 176: Not sure if this should trigger the decorator - why do we want it?

py4DSTEM/io/datastructure/py4dstem/calibration.py
```diff
@@ -178,6 +195,8 @@ def get_origin(self,rx=None,ry=None):
         if any([x is None for x in ans]):
             ans = None
         return ans
+
+    @propagating_calibration
```
bsavitzky, line 199: this also prob should not trigger the decorator

bsavitzky (approved):
as discussed, the decorator definition should be moved to another file

once we add a calibrate method to DataCube, we'll want to get rid of it's internal Q/R pixel setters

bsavitzky:
noooooo my commit broke everythinnnggggg

PR #335, opened 2022-09-13
Update to py4DSTEM.show() defaults

cophus:
Might still need some testing, but these are the low hanging fruit to making .show() easier to use.

One question for debate - when scaling/clipvals are not proviced, should vmin/vmax use sorted relative values or absolute intensities?

PR #343, opened 2022-09-20
Show fix to compute clipvals after masking

tcpekin:
This changes the behavior of `show` to compute `clipvals` after creating a combined mask from the mask that is passed in, and the mask required by the scaling option chosen. 

Old behavior, when data is the `max_dp` from a stack:

image

New behavior:

image

I think this makes a much more sensible default - masked areas shouldn't affect the `vmin` or `vmax` value when plotting. This way, if you pass in a mask, and mask off the central spot, its overwhelming intensity isn't used in the calculation. 

To test that the change didn't break anything, I ran through `show.ipynb` in the `sample_code` directory. In doing so, I updated it and `show.py` to be v0.13.6 compatible.


sezelt (approved):
Appears to work as intended. Just had to resolve a merge conflict because #335 also touched `show.py`

PR #344, opened 2022-09-20
Amorphous branch update

tcpekin:
This catches the amorphous branch up.... it is a huge # of commits, since I did a merge and not a rebase - every time I did a rebase I struggled horribly and the repo was borked. This adds about 40 commits to `dev`.

I welcome any suggestions to fixing this dumpster fire, including starting a new branch or just adding to dev, when it is an even more final state.


sezelt:
My inclination is to just merge this in, since this is a feature branch. When it comes time to merge it with dev I think the diff will end up looking sensible.
I know some people are actually using the amorphous branch as it stands now, so I tagged it at its previous commit in case in case anyone needs to go back to that state.

PR #346, opened 2022-09-22
Adding a MIB reader to py4DSTEM

Quantumstud:
Hi,
In this PR I have added the MIB reader. It is very basic and doesn't have a memory mapping feature. I haven't done pytest on it.


sezelt:
The base branch was set to `main`, so I edited the PR to have it target `dev`

sezelt:
Is there a moderately sized MIB file that we can use for testing this?

bsavitzky:
if there is a small-ish test file, it would be great to put it on google drive, add a key to autoload it with the gdrive downloader, and then add a test function to the test module.  @Quantumstud happy to sit down and do this together if you'd like, it should be fast if you have a small file we can use

bsavitzky (commented):
It would be helpful to have a small sample .mib file to test this on.  Aside from this, everything looks good to me!  I made minor changes to how the cross-file importing is done for consistency.

Quantumstud:
Hi @bsavitzky thank you for your comments. Sure, I can share a MIB file with you. But it is a big MIB file. I will check asap if I have some other MIB files as well.

py4DSTEM/io/import_file.py
```diff
@@ -65,6 +65,8 @@ def import_file(
         data = read_gatan_K2_bin(filepath, mem=mem, binfactor=binfactor, **kwargs)
     # elif filetype == "kitware_counted":
     #    data = read_kitware_counted(filepath, mem, binfactor, metadata=metadata, **kwargs)
+    elif filetype == "mib":
+        data = load_mib(filepath, mem=mem, binfactor=binfactor, scan=(256,256),**kwargs)
```
sezelt, line 69: This seems to hardcode the scan shape to be 256x256, and as I understand it the MIB file does not always have the correct scan shape written inside of it (https://fpdpy.gitlab.io/fpd/convert_merlin_binary.html), so this should be a user-specifiable parameter

Quantumstud: Hi Steve(@sezelt ), 

Yes, Colin also raised this issue previously. I did this to use the general import function of py4DSTEM. 
`dataset = py4DSTEM.import_file(file_data)`
Since no other datasets take the scan value, I didn't know how to only selectively pass the scan value when a mib dataset is loaded. 
For now, the workaround is directly calling the internal loader. For example,
`dataset = py4DSTEM.io.load_mib(file_data,scan=(128,128))`.
This would work for a dataset with size (128,128). 

Please let me know if you have any other ideas I can circumvent this problem.

sezelt: the generic import function `py4DSTEM.import_file` accepts `**kwargs`, which are passed through to the specific readers. So you should just remove the explicit `scan=(256,256)` in this line, and instead within the `load_mib` function get the scan shape from its `**kwargs`. You can have default values as well, by doing something like `scan=kwargs.get("scan",(256,256))`

Quantumstud: Oh right! Yeah. I have made the changes to the code. Shall I create another pull request and close this one?

Quantumstud: Thanks! I have made the changes and pushed it again!

bsavitzky (approved):

PR #347, opened 2022-09-23
calibrate_unit_cell

abruefach:
Introduced calibrate_unit_cell into crystal.py to alter unit cell instead of the braggpeaks. Added a dictionary to look up rules for constraining which lattice parameters/angles are allowed to update based on the point groups - someone should definitely check over those to make sure there are no major issues there. I added the pymatgen imports to support this functionality, but the user can get around this by predefining the 'coef_index' and 'coef_update'.

It is working for polycrystalline gold, I have not checked it for other crystal structures yet.


sezelt:
Do you have an example notebook for testing this?

sezelt:
The new commits pushed into here contain hacks to the automatic calibration function and comment out some asserts in other code that we want to keep.

sezelt:
Do you still plan on reorganizing the additions, breaking off parts of it into a new `crystal_xyz.py` file, like we discussed initially?

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
+
+import numpy as np
+import matplotlib.pyplot as plt
+from fractions import Fraction
+from typing import Union, Optional
+from copy import deepcopy
+from scipy.optimize import curve_fit
+import sys
```
bsavitzky, line 8: `sys` doesn't look like its used, suggest removing

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
+
+import numpy as np
+import matplotlib.pyplot as plt
+from fractions import Fraction
+from typing import Union, Optional
+from copy import deepcopy
```
bsavitzky, line 6: i think using deepcopy is a little dangerous, and should no longer be needed since the BraggVectors.copy() function has been updated

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
+
+import numpy as np
+import matplotlib.pyplot as plt
```
bsavitzky, line 3: `plt` looks like it is unused, suggest removing

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
...
+from fractions import Fraction
```
bsavitzky, line 4: `Fraction` does not appear to be used, suggest removing

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
...
+from copy import deepcopy
+from scipy.optimize import curve_fit
+import sys
+
+from py4DSTEM.io.datastructure import PointList, PointListArray
+from py4DSTEM.process.utils import single_atom_scatter, electron_wavelength_angstrom
+from py4DSTEM.utils.tqdmnd import tqdmnd
+
+from py4DSTEM.process.diffraction.crystal_viz import plot_diffraction_pattern
+from py4DSTEM.process.diffraction.crystal_viz import plot_ring_pattern
```
bsavitzky, line 15: these all appear to be unused, please remove

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
...
+        coefs = (scale_pixel_size, k_broadening, 1.0)
+        bounds = (0.0, np.inf)
+        popt, pcov = curve_fit(fit_profile, k, int_exp, p0=coefs, bounds=bounds)
+
+    scale_pixel_size = popt[0]
+    k_broadening = popt[1]
+    int_scale = np.array(popt[2:])
+
+    # Make a copy of bragg_peaks, apply correction
+    bragg_peaks_cali = deepcopy(bragg_peaks)
```
bsavitzky, line 116: please change to `bragg_peaks_cali = bragg_peaks.copy()`.  This previously had a bug, which i imagine is why deepcopy is being used, but the bug should now be patched, and it would be good to avoid deepcopy on these complex / linked datastructures if possible

py4DSTEM/process/diffraction/crystal_calibrate.py
```diff
@@ -0,0 +1,465 @@
...
+            1.0,
+        )
+        bounds = (0.0, np.inf)
+        popt, pcov = curve_fit(
+            fit_crystal.fitfun, 
+            k, 
+            int_exp, 
+            p0 = coefs,
+            bounds = bounds,
+            )
```
bsavitzky, line 370: It's possible i've just missed something, but: it looks to me like `fit_crystal` is never defined if we enter into this `else:` conditional.  Is that right?  If so this should be fixed

bsavitzky (changes requested):
This looks great.  There are a few changes that would help, all in `crystal_calibrate.py`:
(1) there are a bunch of imports that don't actually get used, and should be removed
(2) it looks like there may be a bug in `calibrate_unit_cell` in the case where `fit_all_intensities` is False, where `fit_crystal` is never defined (though I would think the flaker would catch this and I'm sure you've done some testing, maybe or maybe not of this case, so I'm wondering if I'm just missing something)
(3) please use `braggvectors.copy()` rather than `deepcopy(braggvectors)`.  I recently fixed the bug in `braggvectors.copy()`, and that edit it *not* yet in `dev`, but it should be very shortly - I can send a message to let you know when that's been patched

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -896,6 +876,228 @@ def fit_profile(k, *coefs):
...
+        For example, the default for cubic unit cells is:
+            coef_index = [0, 0, 0, 3, 3, 3]
+            coef_update = [True, True, True, False, False, False]
+        Which allows a, b, and c to update (True in first 3 indices of coef_update)
+        but b and c update based on the value of a (0 in the 1 and 2 positions in coef_index).
+
+        The user has the option to predefine these as well. In the coef_update list, there must
+        be 6 entries and each are boolean. In the coef_index list, there must be 6 entries, with
+        the first 3 entries being between 0 - 2 and the last 3 entries between 3 - 5. These act
+        as pointers to pull the updated parameter from.
```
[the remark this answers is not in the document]
abruefach: I added more details here! Let me know if more is needed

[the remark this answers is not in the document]
abruefach: I added more details here! Let me know if more is needed

bsavitzky (approved):

PR #348, opened 2022-09-23
Bugfix for 'ordered' intensity range

tcpekin:
Honestly this is a one line fix, that prevents a bug from the 'ordered' intensity_range parameter... What's the best way to not bring 5 commits along w/ this fix? Should I try a rebase instead of merge when updating my branch?


sezelt:
It's actually very easy! When you merge a PR it gives you the option to squash it into a single commit.

sezelt (approved):

PR #350, opened 2022-09-26
Introduce featurization - generic clustering class for any numpy array

abruefach:
This class is built to take in any numpy array and perform clustering on it.  A general workflow includes the following (demonstrated using the code and files attached in Featurization_Demo.zip):

1. Instantiating the class with the input features and keys (names) of the features (right now referred to as keys, since they are stored in a dictionary)
2. Prepare on the input features. Features can be concatenated, averaged, scaled, deleted, etc.
3. Call an unsupervised learning method on the features - this currently requires the user to input which features they would like to call the unsupervised learning model on, as well as other input parameters. The method that is best tested is the NMF_iterative. This can be called i number of times (controlled by the iters argument, runs different random seeds) - the output clusters of these models can then be combined to create an ensemble model later on, i.e. 'consensus clustering'
4. Call 'get_class_ims' - this just restructures the results from the unsupervised learning model into real space images.
5. Perform post processing on clusters if required - the demo code I'll send over includes a spatial_separation method, which separates spatially distinct regions based on their x,y coordinates and weighs in the feature maps.
6. Perform consensus clustering - this method performs both label correspondence and consensus clustering, which matches clusters based on similarity and then averages the classes. I may want to split the label correspondence and consensus clustering up in the future, it may make more sense logically for these to be separate.

I'd love suggestions on how to make this module more user friendly/generally better, it is functional right now but not optimal :)

[Featurization_Demo.zip](https://github.com/py4dstem/py4DSTEM/files/9648426/Featurization_Demo.zip)


py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+                    seed = rng.randint(5000)
+                nmf_temp = NMF(n_components = n_components[keys[i]], random_state = seed)
+                nmf_ = nmf_temp.fit_transform(self.features[keys[i]])
+                if nmf_temp.reconstruction_err_ < recon_error:
+                    self.nmf[keys[i]] = nmf_
+                    self.nmf_comps[keys[i]] = nmf_temp.components_
+                    recon_error = nmf_temp.reconstruction_err_
+        return
+    
+    def NMF_iterative(self, keys, max_components, merge_thresh, iters, random_state = None, return_all = True):
```
bsavitzky, line 311: i'd suggest merging this function and the other `NMF` function into a single function.  If we set the `merge_thresh` default to some number greater than one, it should by default reduce to the normal `NMF` function

py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+        gmm, gmm_labels, gmm_proba = {}, {}, {}
+        for i in range(len(keys)):
+            self.gmm[keys[i]], self.gmm_labels[keys[i]], self.gmm_proba[keys[i]] = _gmm_single(self.features[keys[i]],
+                                                                                cv=cv[keys[i]],
+                                                                                components = components[keys[i]],
+                                                                                iters = iters[keys[i]], 
+                                                                                random_state = random_state)
+        return
+
+    def get_class_DPs(self, keys, classification_method, thresh, dc):
```
bsavitzky, line 353: maybe put `dc` at the beginning of the args and name it `datacube`

py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+        for i in range(len(keys)):
+            class_patterns = []
+            if classification_method[keys[i]] == 'nmf':
+                for l in range(self.W[keys[i]].shape[1]):
+                    class_pattern = np.zeros((dc.data.shape[2], dc.data.shape[3]))
+                    x_ = np.where(self.W[keys[i]][:,l] > thresh[keys[i]])[0]
+                    for x in range(x_.shape[0]):
+                        class_pattern += dc.data[x_[x],0] * self.W[keys[i]][x_[x],l]
+                    class_patterns.append((class_pattern - np.min(class_pattern)) 
+                                          / (np.max(class_pattern) - np.min(class_pattern)))
```
bsavitzky, line 372: i'd suggest changing the normalization to divide by the sum of the weights `W[...]`

py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+        return
+
+    def get_class_DPs(self, keys, classification_method, thresh, dc):
+        """
+        Returns weighted class patterns based on classification instance
+        dc (datacube) must be vectorized in real space (shape = (R_Nx * R_Ny, 1, Q_Nx, Q_Ny)
+        
+        Args:
+        keys (list of str): List of strings associated with features
+        classification_method (str): either 'nmf' or 'gmm' - finds location of clusters
```
bsavitzky, line 360: can we have it work for other methods?

py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+            cc_data = []
+            for j in range(len(self.class_ims[keys[k]])):
+                labelled_temp = []
+                for l in range(len(self.class_ims[keys[k]][j])):
+                    image = np.where(self.class_ims[keys[k]][j][l] > threshold, 
+                                    self.class_ims[keys[k]][j][l], 0)
+                    if method == 'yen':
+                        t=threshold_yen(image)
+                    elif method == 'otsu':
+                        t = threshold_otsu(image)
```
bsavitzky, line 437: add a catch for a bad input

bsavitzky (changes requested):
Great start.  Thanks Alex!

py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+        classification_method (str) indicate where to get feature matrix from
+        """
+        for i in range(len(keys)):
+            if classification_method == 'nmf':
+                self.features[keys[i] + '_nmf'] = self.W[keys[i]]
+            elif classification_method == 'pca':
+                self.features[keys[i] + '_pca'] = self.pca[keys[i]]
+            elif classification_method == 'ica':
+                self.features[keys[i] + '_ica'] = self.ica[keys[i]]
+        return
```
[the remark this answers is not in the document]
abruefach: Vestigial method, removed in new version

py4DSTEM/process/classification/featurization.py
```diff
@@ -0,0 +1,637 @@
...
+    def concatenate_features(self, keys, output_key):
+        """
+        Concatenates dataframes in 'key' and saves them to features with key 'output_key'
+        
+        Args
+        keys (list) A list of keys to be concatenated into one array
+        output_key (int, float, str) the key in which the concatenated array will be stored
+        """
+        self.features[output_key] = np.concatenate([self.features[keys[i]] for i in range(len(keys))], axis = 1)
+        return
```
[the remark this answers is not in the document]
abruefach: This is modified in the new version to combine Featurization instances

bsavitzky (approved):

PR #351, opened 2022-09-27
Virtual diffraction update

smribet:
This PR includes: 
 - an update to the `virtual diffraction` module including renaming from `DiffractionImage` to `VirtualDiffraction`
 - bug fixes for `virtual imaging`
 - bug fix for `read` with PC
 
[Here](https://drive.google.com/file/d/1_tC2eIFyy2bHefEScTMdjXLYySt7DdfQ/view?usp=sharing) is a notebook with some examples.


bsavitzky (commented):
Hey @smribet - this looks really good.  I made a number of commits making some updates - would you take a look and let me know if it looks ok to you?  Also, did you have a test file you were using for this new code, and if so, could you try running it with these edits?  I have not tested the changes I made.  If you already have a test file using what you already have it probably best, and if not let's touch base and discuss, and one of us can implement some tests.

I made a few kinds of edits:
(1) docstrings - there were a few things I updated, but the big one was to the whitespace lines and indentation level - that stuff is meaningful in Python!  This is on me, as I have not given clear guidelines for the way to do this in our docstrings - tldr is that there was too much whitespace and too many tabs (except where there were too few ;p)
(2) speaking of tab levels - there were a couple of control flow statements where the indentation level jumped by more than one tab.  I think Python actually does parse this correctly, but it makes me feel sad :O
(3) the actual algorithm in virtualdiffraction.py: I changed this fairly substantially - it's now broken down into 5 different use cases that each get computed a little differently.  I think this should be much faster (in some cases - in some, like whole pattern, it may not matter).  There also was a small error in the mean computation that would lead to incorrect answers in a subset of use cases which should be fixed now.

Let me know what you think whenever you get a chance.  Great work.

py4DSTEM/visualize/show.py
```diff
@@ -936,11 +936,8 @@ def show_rectangles(ar,lims=(0,1,0,1),color='r',fill=True,alpha=0.25,linewidth=2
 
 def show_circles(
     ar,
-    center, 
+    center,
     R,
-    vmin = 0,
```
smribet: I had originally added these changes since the new default options for `show` return results that are a little confusing for `get_probe_size` (examples below with old and updated defaults). I'm realizing this isn't the best place for this change. Instead I added a plotting option with different defaults for `datacube_fns.` What do you think?
Screen Shot 2022-10-05 at 11 44 47 AM
Screen Shot 2022-10-05 at 12 00 05 PM

bsavitzky:
Hey @smribet - ok, I've pushed my latest updates to this branch!  I'll send you a test file on slack

bsavitzky (approved):

PR #352, opened 2022-09-28
Various small fixes, mostly in ACOM

cophus:


cophus:
Note that my commenting out of the calibration stuff might be because I was using autoreload magic - feel free to revert those changes if needed.

sezelt:
I pushed some edits to clean things up, and also potentially fix the problem with `Calibration` objects that are loaded from H5 files. Can you re-test that notebook where the `Calibration` bug came up because of the calibration being loaded from file? I don't have a good test file for that on hand...

py4DSTEM/io/datastructure/py4dstem/io.py
```diff
@@ -42,14 +42,11 @@ def Calibration_from_Metadata(metadata):
...
     """
     from py4DSTEM.io.datastructure.py4dstem.calibration import Calibration
-    p = metadata._params
-    metadata.__class__ = Calibration
-    metadata.__init__(
-        name = metadata.name
-    )
-    metadata._params.update(p)
+    
+    cal = Calibration(name = metadata.name)
```
sezelt, line 46: This should fix the calibration problems, where `_targets` was missing, but I want a test of that before merging

sezelt (commented):
All the ACOM changes look right, but I want to fix the Calibration problems the right way before merging. So if you can just test my tweaks then we should be set

bsavitzky (approved):

PR #356, opened 2022-11-01
Parallelizing the dynamical structure factor calculations

cophus:
SF calculations currently take a very long time for the dynamical case - this is my attempt to parallelize it.


sezelt:
Closing this as #361 supersedes it

PR #357, opened 2022-11-02
Added workflow to automate publishing to pypi

Quantumstud:
Hi,
I have added a GitHub workflow that automatically checks the versioning and increments the patch number if the version hasn't changed. It would then push the new version to PyPI.

Thanks!


.github/workflows/pypi_upload.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Get changed files
+        id: changed-files-specific
+        uses: tj-actions/changed-files@v32
+        with:
+          files: |
+            py4DSTEM/version.py
+      - name: Run step if any file(s) in the docs folder change
+        if: steps.changed-files-specific.outputs.any_changed == 'false'
+        run: |
+          echo "Version file not changed, run script to change the version file."
```
bsavitzky, line 29: suggest changing to "running" rather than "run" - this happens automatically, but "run" sounds like the user needs to do something

.github/workflows/pypi_upload.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install build
+      - name: Build package
+        run: python -m build
+      - name: Publish package
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_TOKEN }}
```
bsavitzky, line 51: I named this token "PYPI_API_TOKEN" in the repo secrets, per your instructions on slack - should this be identical here?

bsavitzky (changes requested):
This looks great!  Thanks for this PR, this will really help our distribution.

I made two small comments on the .yml file, please see those.  Additionally, there was one more thing I was hoping we could get this GH action to do: after you check/update the version, but before you push to PyPI, is it possible to pull/merge `main` into `dev`?  That way every time we push in to dev it syncs up these branches and we can continue working seemlessly on `dev` without being a version (and potentially other code) behind.  I realize this may be more complicated....hopefully it's relatively easy to do as long as there are no merge conflicts...?

.github/workflows/pypi_upload.yml
Quantumstud: Hi Ben,
This is a restricted command for the GitHub workflow, which we cannot change. This instructs the workflow to run a specific set of instructions while executing the Github actions.

PR #361, opened 2022-11-11
Optimize dynamical scatting factor calculation

sezelt:
Using `mp-1247858` with `k_max=2.0` for testing, the old method took about 8 minutes to compute the WK-CP scattering factors. @cophus got this down to 1 minute and 12 seconds in PR #356. This further optimized method takes only 6 seconds. For simpler structures it is also way faster than it used to be: dc-Si with `k_max=4.0` used to take 2 seconds, now it's 0.1 seconds. 

I was not able to push my changes into #356 so I am just making this a new PR, but it is based on the same branch. So I will also close #356 as a result


sezelt:
Also adding an option to round off the _g_ vector magnitudes for computing the atomic form factors so that the caching does not get messed up by floating point errors, and to allow some nice speedup at the cost of some precision if the user wants to go fast.

cophus (approved):
looks good to me!

PR #364, opened 2022-12-05
Allowing py4DSTEM.show() to combine multple arrays into a single image

cophus:
with the new argument:
combine_images=True


cophus:
Here is an example of the new syntax:

image

smribet (approved):

PR #371, opened 2022-12-22
DPC/Ptychography Additional Constraints, User-Interface Tweaks

gvarnavi:
This PR introduces additional constraints and various user-interface tweaks to the DPC/Ptychography modules which came up as @smribet and @cophus were using it on experimental data/tutorial notebooks respectively.

Object constraints:
- Smoothness constraint (Gaussian filtering)
- Threshold constraint (amplitude <1)
  - Accepts a `pure_phase_object` flag to set amplitude to 1

Probe constraints:
- Center-of-mass centering
- Fourier amplitude fixing (useful for measurements with a vacuum-probe)
- Finite support constraint (supergaussian)

Additionally, this introduces preliminary probe-position correction:
- Triggered by specifying `fix_positions_iter` < `max_iter`
- Default uses gradient descent on measured/estimated exit waves difference
- Specifying `alternative_position_correction=True` instead uses a least-squares approach on measured intensities

User-interface improvements:
- `DPCReconstruction.reconstruct()` and `PtychographicReconstruction.reconstruct()`  now accept a `reset` flag to restart reconstructions. Default is False
- `plot_center_of_mass` argument of `.preprocess()` methods now accepts a string. When set to 'default', only the corrected CoM arrays are plotted. If set to 'all', the raw, normalized, and corrected arrays are plotted. 
-  Added `warnings.simplefilter(action="always",category=UserWarning)` to ensure warnings get printed when user runs same jupyter notebook more than once


sezelt:
It looks like most? all? of the commits here were bundled up in #372, but because I squashed it on merge I think github is not detecting that these PRs do related things. So I am confused why it still shows such a large diff...

gvarnavi:
Yes, I was just checking this. Indeed, seems like the merged #372 makes this PR redundant.

As a side-note though,it seems like scipy.ndimage.gaussian_filter only recently started supporting complex arrays, so this might be an issue..

gvarnavi:
Merged #372 before the last push, so the diff is more reasonable now.
Flight tweaks include the following:
- fixed tqdnm warning bug
- added `.pad_Q()` and `.resample_Q()` datacube methods
- updated ptychography module to use these for more flexibility in setting reciprocal-space and real-space sampling
- quick fix for the [now expired](https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations) `np.float` and `np.int` aliases for numpy 1.24

bsavitzky (approved):

PR #372, opened 2022-12-22
Small fixes for tutorial notebooks, some of the phase contrast updates

cophus:
Includes some updates for the upcoming phase contrast PR for compatbility


sezelt (approved):
I'm giving this a "skimming it while sitting at the gate for a delayed flight" kind of approval, so to speak. 
I'm curious what the `warnings.simplefilter` is there for in a few files. If it's related to preventing `print`ed messages from messing with the `tqdm` progress bar, the more elegant way to do this is to use the `write` method of the progress bar. But if it's a warning that's getting printed, I'm not sure there is a better way that isn't overly complex.

gvarnavi:
Here's a "didn't know SFO international terminal had an outdoor terrace with WiFi reception, while I wait for my flight" kind of reply then:

This is unrelated to the `tqdm` issue @cophus mentioned on slack, but rather it's to ensure warnings get printed twice if a user runs the same cell. I believe default behavior in a notebook is for them to only get printed on the first call: https://docs.python.org/3/library/warnings.html#the-warnings-filter

PR #377, opened 2023-01-20
Adding transpose and flip to .ang orientation map saving

cophus:
Two new options:
    transpose_xy = True,
    flip_x = True,
This allows for consistent output when user wants x axis to be along horizontal direction.


cophus:
I also suppressed the gimbal lock warning.  It's fine for orientation maps, and very tedious to dump thousands of warnings onto the console.

bsavitzky (approved):

PR #387, opened 2023-01-25
Add elliptical resampling

sezelt:
In some analyses (e.g. ptychography) it is necessary to correct distortions of the raw diffraction patterns. This PR adds a function based on `cartesian_to_polarelliptical_transform` called `elliptical_resample` that resamples a diffraction pattern to correct elliptical distortion while remaining in Cartesian coordinates. There is also a utility function that loops over every diffraction pattern in a DataCube and applies the resampling operation.


sezelt:
This notebook shows an example usage based on a toy test object. 

[elliptic resampling.ipynb.zip](https://github.com/py4dstem/py4DSTEM/files/10502767/elliptic.resampling.ipynb.zip)

bsavitzky (approved):

PR #389, opened 2023-01-26
EMPAD manual shape override

sezelt:
Many EMPAD files don't have the data shape in their metadata, so this adds a manual way to specify the shape for loading. At some point we can implement an automatic method, which may take the shape of:
* Look at the file size and make an educated guess (arbitrary shapes are not possible, I think, so it's always 64x64, 128x128, 256x256)
* Run a regular expression on the file name and hope it's written according to the standard rules
* Search for the XML file containing the metadata that is sometimes present


sezelt:
After thinking about it for a hot second, I'm going to just go with the filesize based method for automatic detection. I'm also leaving the manual override just in case, though

bsavitzky (approved):

PR #390, opened 2023-01-30
image grid and Bragg vector calibration fixes

smribet:
This PR has: 
(1) a quick fix for strain mapping for when the braggvectors are calibrated
(2) some options for `show_image_grid` to customize a title and return the figure. For example: 
```
fig, ax = plt.subplots(1,2, figsize = (5,3))
py4DSTEM.show(
    [dc.data[0,0], 
     dc.data[0,1]], 
    figax = (fig,ax), 
    title = ['ducks',"more ducks"],
    title_index = True, 
    suptitle = "animals",
    get_bordercolor=lambda i:['red','blue'][i],
)

ax[0].scatter(56,56, s = 160, color = 'lime')
```
attachment

### DIFF
--- a/py4DSTEM/io/datastructure/py4dstem/braggvectors_fns.py
+++ b/py4DSTEM/io/datastructure/py4dstem/braggvectors_fns.py
@@ -320,7 +320,8 @@ def index_bragg_directions(
     """
     From an origin (x0,y0), a set of reciprocal lattice vectors gx,gy, and an pair of
     lattice vectors g1=(g1x,g1y), g2=(g2x,g2y), find the indices (h,k) of all the
-    reciprocal lattice directions.
+    reciprocal lattice directions. In units of pixels. If calibrated in A^-1, also stores
+    `self.braggdirections_calibrated.` 
 
     Args:
         x0 (float): x-coord of origin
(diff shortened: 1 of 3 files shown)


py4DSTEM/io/datastructure/py4dstem/braggvectors_fns.py
```diff
@@ -385,13 +399,22 @@ def add_indices_to_braggpeaks(
...
-        self.braggdirections,
-        maxPeakSpacing = maxPeakSpacing,
-        qx_shift = self.Qshape[0]/2,
-        qy_shift = self.Qshape[1]/2,
-    )
+    if self.calibration.get_Q_pixel_units() == 'A^-1':
+        bragg_peaks_indexed = add_indices_to_braggpeaks(
+            self.vectors,
+            self.braggdirections_calibrated,
+            maxPeakSpacing = maxPeakSpacing * self.calibration.get_Q_pixel_size(),
```
sezelt, line 406: Would it make more sense for `maxPeakSpacing` to be in inverse angstroms in the case that the data is calibrated? Keeping it in original pixel units could be confusing

py4DSTEM/io/datastructure/py4dstem/braggvectors_fns.py
```diff
@@ -345,6 +346,19 @@ def index_bragg_directions(
 
     self.braggdirections = braggdirections
 
+    if self.calibration.get_Q_pixel_units() == 'A^-1':
```
sezelt, line 349: I like that we are taking the hardline atomic physicist approach, that the data isn't calibrated unless it's calibrated in Ångström units. SI is for _chemists_

sezelt (approved):

PR #395, opened 2023-02-07
changes to crystal phase

Quantumstud:
Introduced counting the false positive in the matrix before the NNLS algorithm in the crystal phase class.


Quantumstud:
I am closing the pull request since this Colin made a better code which works well.

PR #410, opened 2023-04-14
convert 1/nm in DM to A^-1

sezelt:
This is a simple fix for the DM reader that checks if the pixel size is "1/nm" and converts it to "A^-1".


cophus (commented):
Hmmm not sure why I thought it would be more difficult than this!   I guess the only question is - do we want to auto convert to 1/A, or allow 1/nm?

I also see in this code snippet that "lambda" is misspelled right above.

sezelt:
My preference would be to make stuff be in inverse angstroms as much as possible. In ptycho, for example, there is no handling of "1/nm", only "A^-1".

bsavitzky (approved):
Thanks Steve :D

PR #423, opened 2023-05-03
bugfix in virutalimage.py and get_virtual_image speedup

tcpekin:
This first fixes a bug in dask I found in `virtualimage`, and then speeds up the case where the center does not need to be shifted by quite a lot.

image

Two differences - no more tqdm for the virtual image in the case where it doesn't loop (non-shifted center), and user defined masks are converted to the `bool` dtype, as when they were ints, it was blowing up my computer memory and not actually running.

See PDF benchmark
[benchmark.pdf](https://github.com/py4dstem/py4DSTEM/files/11391113/benchmark.pdf)

### DIFF
--- a/py4DSTEM/process/virtualimage.py
+++ b/py4DSTEM/process/virtualimage.py
@@ -108,11 +108,12 @@ def get_virtual_image(
 
             # set up a generalized universal function for dask distribution
             def _apply_mask_dask(datacube,mask):
-                virtual_image = np.sum(np.multiply(datacube.data,mask), dtype=np.float64)
-            apply_mask_dask = da.as_gufunc(
+                return np.sum(np.multiply(datacube.data,mask), dtype=np.float64)
+            apply_mask_dask = da.gufunc(
                 _apply_mask_dask,signature='(i,j),(i,j)->()',
                 output_dtypes=np.float64,
                 axes=[(2,3),(0,1),()],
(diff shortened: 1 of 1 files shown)


tcpekin:
Also, I am sure my dask bugfix is not the best way to do it, with the `allow_rechunk` option, but otherwise it raised an error.

sezelt:
Nice catch on the dask bug!

I think we do not want to restrict ourselves to boolean masks, having a mask with soft edges or one with some positive and some negative regions for a differential image is useful. 

As for performance, vectorizing the mask application by either broadcasting-multiplying or masking is indeed more efficient for very small memory-resident arrays, but in other cases the performance is far worse. In the past we have used a variety of ways of computing the virtual images, and we ended up removing them because they failed in so many cases.

Perhaps we should consider implementing some kind of `is_array_small` utility that tests that an array is a plain `np.ndarray` (not a memmap or an h5 Dataset or anything else exotic) and that it is much smaller than the available RAM. Then if those conditions are met we can use a vectorized operation, and otherwise stick to the for loop.

tcpekin:
Makes sense to me, I didn't think that something like this would have really slipped past you guys.

sezelt:
This seems to have been made stale by v14 changes to the virtual imaging functions so I'm going to close it

PR #428, opened 2023-05-12
datacube copy

smribet:


bsavitzky (changes requested):
This should almost do it! The dim vectors also need to be transferred.  I would suggest doing this using the calibration that you've already copied

smribet:
Good catch, thanks Ben!  Is this how you would suggest updating the dim vectors?

bsavitzky (approved):

PR #429, opened 2023-05-12
Error is now stored even if store_iterations=False. 

hisigns:
Changed in singleslice, multislice, mixedstate, simultaneous, overlap, and overlap_magnetic.


gvarnavi:
Thanks Hannah,
Seems like `self.error_iterations = []` needs to initialized in all cases except when user explicitly sets `reset=False`. Otherwise looks goods, merged and tweaked upstream!

PR #431, opened 2023-05-16
Fix for show() default behaviour in sparse images

cophus:
Error occurs with mostly-white or mostly-black images when vmin==vmax (the automatically determined values for vmin and vmax)


cophus:
Examples

old behaviour:
image

new behaviour:
image

bsavitzky (approved):

PR #432, opened 2023-05-16
Adding intensity scaling to whole pattern fitting

cophus:
We found that not scaling the DP intensities causes the fits to be poor and take a long time.  Dividing each pattern by the mean intensity before fitting fixes both of these issues.


cophus:
Fixing plotting bugs in show() too:
image

alex-rakowski:
Not sure how, but I have committed to this PR... I think I need to work with you @cophus  to remove them

alex-rakowski:
I should have removed all the weird commits let me know if I made any other errors I didn't spot

PR #439, opened 2023-05-18
Bayesian Optimization with Gaussian Process for Ptycho

sezelt:
This adds a relatively friendly way to optimize the hyperparameters used in ptychography reconstructions. We accomplish this through the `PtychographyOptimizer` class, which accepts a dictionary that specifies both the fixed and the variable hyperparameters. The optimizer then uses the gaussian process minimizer in scikit-optimize to find the parameters that give the lowest reconstruction error. 

Potential Todos:
- [x] Documentation!
- [x] A way to refine the reciprocal pixel size
- [x] A way to refine the x and y realspace pixel scales together rather than via the affine transform
- [x] Decide where to put dependencies
- [x] Do not re-run preprocess if there are no optimization variables up to that point


sezelt:
Example usage is as follows:
```python 
from py4DSTEM.process.phase.parameter_optimize import OptimizationParameter, PtychographyOptimizer
init_args = {
    'datacube':dataset,
    'energy':80e3,
    'semiangle_cutoff':semiangle_cutoff,
    'defocus':OptimizationParameter(-100.,100.,50.),
}
affine_args = {
    'scale0':OptimizationParameter(0.9,1.1,1.0),
    'scale1':OptimizationParameter(0.9,1.1,1.0),
}
preprocess_args = {
    'force_com_rotation':rotation_degrees,
    'force_com_transpose':False,
    'fit_function':'constant',
}
recon_args = {
    'max_iter':8,
    'step_size':0.5,
    'gaussian_filter_sigma':1.0,
    'fix_probe_iter':np.inf,
}
optimizer = PtychographyOptimizer(
    py4DSTEM.process.phase.SingleslicePtychographicReconstruction,
    init_args = init_args,
    affine_args = affine_args,
    preprocess_args = preprocess_args,
    reconstruction_args = recon_args,
)

optimizer.optimize(
    n_calls=20,
    n_initial_points=1
)
optimizer.visualize()
```

gvarnavi:
Thanks Steve, looks great - excited to try it!

Quick question :
- It seems like init/preprocess happens regardless of if those parameters are all static? Like say I only had `OptimizationParameter` inside `recon_args`, is there a way to skip init/preprocess and simply run the optimizer using the `reset=True` parameter of `reconstruct()`?
  - Example use-cases would be optimizing the (a,b,c) `reconstruction_parameter` in the generalized projection sets, or optimizing various regularization constraints

Some thoughts on to-dos:
- Refining reciprocal pixel size: Indeed, this would require either modifying the datacube calibration or a patch to overwrite the `_angular_sampling` and `_reciprocal_sampling` attributes after the `_extract_intensities_and_calibrations_from_datacube()` call inside `preprocess()`.

- Refining real pixel size: Perhaps a simple `scale_both` or similar argument inside `affine_args` which sets both `scale0` and `scale1` in the appropriate `AffineTransform` calls.

- It would also be nice to have read/write functionality for the optimizer (or equivalently, have a way of attaching the optimizer to the ptycho class and then modify read/write in those classes).

sezelt:
* At the moment, yes it recreates the ptycho object each time rather than running reset. It should not be too hard to flag it to hold on to the same object, but do you think it is worth it performance-wise? 
* Are there side effects that need to be taken care of if I change `_angular_sampling` or `_reciprocal_sampling` after `preprocess`ing?
* Agreed. 
* Saving the result would be interesting, so that you could save the GP model and the convergence, but it might be hard to adapt the scikit-optimize structure into a py4DSTEM object.

gvarnavi:
Quick answers:
- Yes, I think it would be worth it performance-wise to avoid preprocessing, especially for datasets w/ a lot of probe positions.
- `_angular_sampling` (together with `_region_of_interest_shape`) controls `sampling` and thus things like the size of the object, so you'll need to change it right after `_extract_intensities_and_calibrations_from_datacube()`. Perhaps we add a `force_angular_sampling` flag to `preprocess`.
- Re: saving - yeah, I'd mostly be interested in the saving (some of) the sampled parameter space and the resulting reconstructions.

bsavitzky:
> * Saving the result would be interesting, so that you could save the GP model and the convergence, but it might be hard to adapt the scikit-optimize structure into a py4DSTEM object.

I think the question is - what do you want to save?  Some function like e.g. `ptycho_reconstruction.store_optimizer(ptycho_optimizer)` could alter the substructure s.t. at save time, you would store the argument dictionaries that built the optimizer, the arguments passed to the optimizer, and the converged values.  That should not be hard!  @gvarnavi it sounds like maybe you want to save more than this - you mean some subset of the sampled values and their reconstruction errors?  Or are you saying you would want some subset of the reconstructions as well?

sezelt:
I've also moved the pixel size re-calibration from the `ptycho` function into the `DataCube` one, so that when I resize a dataset outside of a ptychography run, the pixel size is updated appropriately.

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+        n_iterations: int,
+        init_static_args: dict,
+        affine_static_args: dict,
+        preprocess_static_args: dict,
+        reconstruct_static_args: dict,
+        init_optimization_params: dict,
+        affine_optimization_params: dict,
+        preprocess_optimization_params: dict,
+        reconstruct_optimization_params: dict,
+    ):
```
gvarnavi, line 286: Adding flags for error divergence and log error.

```suggestion
        check_error_divergence: bool,
        return_log_error:bool,
    ):
```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+            # Apply affine transform to pixel grid, using the
+            # calibrations lifted from the dataset
+            dataset = init_static_args["datacube"]
+            init_args["initial_scan_positions"] = self._get_scan_positions(tr, dataset)
+
+            ptycho = obj(**init_args)
+            prep(ptycho, **prep_args)
+            recon(ptycho, **reco_args)
+
+            return np.log(ptycho.error)
```
gvarnavi, line 353: Error divergence check and log error suggestion:

```suggestion
            errors = ptycho.error_iterations
            if check_error_divergence and errors[-1] > min(errors[:-1]):
                error = 1.0
            else:
                error = errors[-1]
            
            if return_log_error:
                error = np.log(error)
            
            return error
```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+            n_calls,
+            self._init_static_args,
+            self._affine_static_args,
+            self._preprocess_static_args,
+            self._reconstruction_static_args,
+            self._init_optimize_args,
+            self._affine_optimize_args,
+            self._preprocess_optimize_args,
+            self._reconstruction_optimize_args,
+        )
```
gvarnavi, line 140: Error divergence check and log error suggestion (see below):

```suggestion
            check_error_divergence = check_error_divergence,
            return_log_error = return_log_error,
        )
```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+
+        self._reconstruction_type = reconstruction_type
+
+        self._set_optimizer_defaults()
+
+    def optimize(
+        self,
+        n_calls: int,
+        n_initial_points: int,
+        **skopt_kwargs: dict,
```
gvarnavi, line 112: Error divergence check and log error suggestion (see below):

```suggestion
        check_error_divergence = True,
        return_log_error = True,
        **skopt_kwargs: dict,
```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+            List of options for Categorical parameter
+        """
+        # Check input
+        space = space.lower()
+        assert space in (
+            "real",
+            "integer",
+            "bool",
+            "categorical",
+        ), f"Unknown Parameter type: {space}"
```
gvarnavi, line 409: Assert is really for debugging, not catching user ValueErrors in my book (?)

```suggestion
        if space not in ("real", "integer", "bool", "categorical"):
            raise ValueError(f"Unknown Parameter type: {space}")

```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+
+        scaling = scaling.lower()
+        assert scaling in ("uniform", "log-uniform"), f"Unknown scaling: {scaling}"
```
gvarnavi, line 412: Assert is really for debugging, not catching user ValueErrors in my book (?)

```suggestion
        if scaling not in ("uniform", "log-uniform"):
            raise ValueError(f"Unknown scaling: {scaling}")
```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+        self._categories = categories
+        self._lower_bound = lower_bound
+        self._upper_bound = upper_bound
+        self._scaling = scaling
+        self._param_type = param
+
+    def _get(self, name):
+        self._name = name
+        if self._param_type is Categorical:
+            assert self._categories is not None, "Empty list of categories!"
```
gvarnavi, line 438: See above. 
```suggestion
```

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+            "integer": Integer,
+            "bool": Categorical,
+            "categorical": Categorical,
+        }
+        param = space_map[space]
+
+        # If a boolean property, the categories are True/False
+        if space == "bool":
+            categories = [True, False]
+
```
gvarnavi, line 426: Assert is really for debugging, not catching user ValueErrors in my book (?)

```suggestion

        if not categories:
            raise ValueError("Empty list of categories!")
```

gvarnavi (commented):
Ok, just got around to review this - pardon the delay.
Great job, looks amazing!

Briefly the comments/suggestions are:
- visualization updates to look nicer, add some more customization, and always return a single fig
- some docstring clarifications
- some stylistic suggestions (namely not using `assert`, and an alternative way to update the optimized dicts using the merge operator)
- error divergence and log error flag suggestions

py4DSTEM/process/phase/iterative_mixedstate_ptychography.py
```diff
@@ -52,7 +52,9 @@ class MixedstatePtychographicReconstruction(PtychographicReconstruction):
     num_probes: int, optional
         Number of mixed-state probes
     semiangle_cutoff: float, optional
-        Semiangle cutoff for the initial probe guess
+        Semiangle cutoff for the initial probe guess in mrad
+    semiangle_cutoff_pixels: float, optional
+        Semiangle cutoff for the initial probe guess in pixels
     rolloff: float, optional
         Semiangle rolloff for the initial probe guess
```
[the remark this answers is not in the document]
sezelt: The reason for this being a pixel parameter is for when you are trying to refine the reciprocal pixel size. You can very easily determine the size of the direct beam in pixels, and this is a constant regardless of rescaling of the reciprocal pixel size. 
The rolloff should probably remain fixed relative to the calibrated semiangle, not the pixel value, but I am not certain about this. I may try optimizing over the rolloff value just to investigate the sensitivity.

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+    """
+    Optimize ptychographic hyperparameters with Bayesian Optimization of a
+    Gaussian process. Any of the arguments to the ptychographic
+    init-preprocess-reconstruct pipeline can be optimized over (including
+    boolean or other flag options)
+    """
+
+    def __init__(
+        self,
+        reconstruction_type: type[PhaseReconstruction],
```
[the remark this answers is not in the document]
sezelt: This is correct as written, so long as the Python version is >= 3.9. See [here](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#the-type-of-class-objects). The optimizer takes the _class_ as an argument, not an _instance_ of the class.
> Sometimes you want to talk about class objects that inherit from a given class. This can be spelled as type[C] (or, on Python 3.8 and lower, [typing.Type[C]](https://docs.python.org/3/library/typing.html#typing.Type)) where C is a class. In other words, when C is the name of a class, using C to annotate an argument declares that the argument is an instance of C (or of a subclass of C), but using type[C] as an argument annotation declares that the argument is a class object deriving from C (or C itself).

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+            np.arange(dataset.R_Ny) * R_pixel_size,
+        )
+        x, y = np.meshgrid(x, y, indexing="ij")
+        scan_positions = np.stack((x.ravel(), y.ravel()), axis=1)
+        scan_positions = scan_positions @ affine_transform.asarray()
+        return scan_positions
+
+    def _get_optimization_function(
+        self,
+        cls: type[PhaseReconstruction],
```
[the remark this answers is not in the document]
sezelt: Same as above, it is correct as written as the function expects a class object, not a class instance

py4DSTEM/process/phase/parameter_optimize.py
sezelt: This will be handed in a separate and larger commit that I'll merge with the PR review commits and push later

sezelt:
I've accepted some of the suggested changes, and pushed my new way of accepting custom error metrics. I'll continue to work on the other suggestions and then run a final test that everything plays nicely together (which will also probably become a candidate tutorial notebook)

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+        """
+        self._init_static_args["verbose"] = False
+
+        self._preprocess_static_args["plot_center_of_mass"] = False
+        self._preprocess_static_args["plot_rotation"] = False
+        self._preprocess_static_args["plot_probe_overlaps"] = False
+
+        self._reconstruction_static_args["progress_bar"] = False
+        self._reconstruction_static_args["store_iterations"] = False
+        self._reconstruction_static_args["reset"] = True
```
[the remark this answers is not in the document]
sezelt: My thinking is that we probably should not spit out plots on every run, since that will undoubtedly crowd up a notebook. It would not make a lot of sense unless we also had it print out the parameters at each iteration as well, which would need some flags or callbacks or something. Probably the best thing for a user to do is to explore the limits of how much defocus they can handle manually before running the optimizer. Alternatively, if you were to come up with some kind of metric that determines if the probe fits well inside the box, we could have it short-circuit and return the max error if the ptycho is run with parameters that can't fit.

py4DSTEM/process/phase/parameter_optimize.py
```diff
@@ -0,0 +1,449 @@
...
+from skopt import gp_minimize
+from skopt.space import Real, Integer, Categorical
+from skopt.plots import (
+    plot_convergence,
+    plot_gaussian_process,
+    plot_evaluations,
+    plot_objective,
+)
+from skopt.utils import use_named_args
+import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
sezelt: oops this one was hidden, i ended up adding it separately

py4DSTEM/process/phase/parameter_optimize.py
sezelt: handled by the new custom error metrics

py4DSTEM/process/phase/parameter_optimize.py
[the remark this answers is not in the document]
sezelt: OK yeah let's add it

py4DSTEM/process/phase/iterative_mixedstate_ptychography.py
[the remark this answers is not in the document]
gvarnavi: Can we resolve this? I think it's fine to keep rolloff in mrad for now.

gvarnavi:
@sezelt Seems like we're all set here? I'll go ahead and merge into `phase_contrast` tomorrow morning if so.

sezelt:
Yes, ready to go

PR #441, opened 2023-05-29
Fix dev install setup.py

rpsankaran:
python version needs to be more permissive

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
     author_email='ben.savitzky@gmail.com',
     license='GNU GPLv3',
     keywords="STEM 4DSTEM",
-    python_requires='>=3.9,<3.11',
+    python_requires='>=3.9,<=3.11.3',
     install_requires=[
         'numpy >= 1.19',
         'scipy >= 1.5.2',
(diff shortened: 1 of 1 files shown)


sezelt:
Earlier we had some discussion on Slack and concluded that we wanted to set the maximum version to 3.10. However, it seems that the reasons for that are no longer a hindrance so I think it is ok to also allow 3.11 now. I am not sure that `<=3.11.3` is the right cutoff, though... I'll investigate further and we can update this.

sezelt:
The build is failing with weird `emdfile` errors: @bsavitzky do we need to bump the version for that? It's at `>=0.0.6` at the moment. 

Also, I think it would be good to sync the Python version specs in the build test scripts to match this change

bsavitzky:
Hey @rpsankaran - thanks for this PR.  Could you rebase your branch with dev? You PR is failing checks due to a bug that's been patched in the dev branch, so if you pull that into your branch you should pass all checks. Then we can merge your PR.

rpsankaran:
> 

HI Ben,

Sorry I'm just seeing this now.  I'm getting up to speed on py4DSTEM first now, going through tutorials, then I'll get into development as I start maybe breaking things :)

PR #443, opened 2023-06-03
backtracking in DPC

sezelt:
This adds a simple backtracking step to the iterative DPC, which rejects steps that increase the error.

### DIFF
--- a/py4DSTEM/process/phase/iterative_dpc.py
+++ b/py4DSTEM/process/phase/iterative_dpc.py
@@ -420,9 +420,6 @@ class DPCReconstruction(PhaseReconstruction):
             xp.mean(self._com_x.ravel() ** 2 + self._com_y.ravel() ** 2)
         )
 
-        if new_error > error:
-            step_size /= 2
-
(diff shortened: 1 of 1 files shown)


gvarnavi:
Hey @sezelt - this looks good, but perhaps let's move the actual step halving outside `forward` and in the same back-tracking if statement in the main loop to keep things cleaner?

Also, if we add the functionality to store the previous iteration to the ptycho classes too then we could also easily implement [Nesterov acceleration](https://jlmelville.github.io/mize/nesterov.html). Happy to take a crack at this.

smribet (approved):
Looks good to me. Thanks @sezelt and @gvarnavi!

gvarnavi:
Quick note on the momentum-acceleration idea referenced above. Implemented this for single-slice and then decided it's not worth supporting for all the classes and different reconstruct parameters (specifically `max_batch_size` and `reset`). 

Kept the small recursive prefactor factors & a reference link in the utils file in commit 445cca1, for possible re-implementation after restructuring.

PR #459, opened 2023-06-22
Rm rooted node

bsavitzky:
- removes the RootedNode class
- bumps emdfile version to 0.0.9
- updates / fixes google drive downloader
- adds test data command-line download tool


sezelt (approved):
Everything looks fine to me... I haven't tested the `RootedNode` removal but I trust it will not break things

PR #461, opened 2023-06-27
Oddly, not all datasets have even dimensions, and we wouldn't want to cut corners

gvarnavi:
This PR mostly ensures odd-dimension DPs get handled correctly:  
- This is achieved by storing all probe/overlap related functions as top-left corner centered arrays
- All "public"-facing attributes such as `self.probe` as-well as all visualization functions center the arrays before returning to the user
- This also means that the read/write functionality changed somewhat and is incompatible for 0.14.1 reconstructions stored w/ iterations (I suspect this affects very few users other than the dev team, but we could consider patching if need be)

Additionally, the PR introduces the following minor changes:
- DPC backtracking and anti-gridding filter
- Switches to svd-like probe orthogonalization for mixed-state
- Re-introduces probe aperture fixing
- Supports `((left_x,right_x), (left_y,right_y))` padding format
- Supports `(global_com_x,global_com_y)` CoM shift format
- Adds `show_complex` to top-level imports
- Comments out a leftover debug print statement in the legacy Datacube readers

~The PR is in a draft state until #439 is merged into `phase_contrast`.~


smribet (approved):
Good stuff!

PR #464, opened 2023-06-29
Update whole pattern fitting

bsavitzky:
updates the whole_pattern_fitting branch to the current dev branch


sezelt (approved):
It looks like this is doing all the right things with regards to WPF (i.e. pulling the latest features, which exist in dev, back to this branch). Hopefully the emdfile-related changes don't break anything in WPF, but I suspect it's minor at worst

bsavitzky:
The emdfile stuff shouldn't break anything for now - I pinned the emdfile version in the setup.py to v0.0.8 for now, which avoids the weird RootedNode issues.  I'll loosen that in dev once everything is ready for it, I think we can get it ready for the next version update

PR #467, opened 2023-07-10
Adding constant probe amplitude regularization function

cophus:
This function fits a step edge model to the probe in polar coordinates, and by default enforces constant intensity inside the probe aperture.  It can be called with:

probe_corr = py4DSTEM.process.phase.utils.regularize_probe_amp(
    probe_init,
    width_max_pixels = 3.0,
)

at a minimum.  Probe phase is not changed.


gvarnavi:
Looks good, will merge and clean-up in follow-up commit to `phase_contrast` as-opposed to using suggestions.

PR #469, opened 2023-07-11
MM2023 tutorials

smribet:
Some bug fixes for the M&M tutorials:

- [x] orient_strain_01_WS2
- [x] diffraction_01_kinematical
- [x] diffraction_02_dynamical
- [ ] orientation_01_AuAgPd_wire.ipynb


py4DSTEM/preprocess/preprocess.py
```diff
@@ -281,6 +281,12 @@ def bin_data_diffraction(
     # set calibration pixel size
     datacube.calibration.set_Q_pixel_size(Qpixsize)
 
+    # remake Cartesian coordinate system
+    datacube._qyy,datacube._qxx = np.meshgrid(
+        np.arange(0,datacube.Q_Ny),
+        np.arange(0,datacube.Q_Nx)
+    )
+
```
bsavitzky, line 289: the `._qxx,._qyy` properies are set in datacube.calibrate - we shouldn't overwrite them here

bsavitzky (changes requested):
This looks good.  I found one bit of code that looks incorrect in preprocess, and there appears to be some merge conflicts (I suspect it might be the same bit of code, I haven't checked yet).  As long as we agree on correct behavior there, I can edit that and then approve

bsavitzky (approved):

PR #483, opened 2023-07-17
more bvects copy bugfixes

bsavitzky:


cophus (approved):
Fixes tested in flowline orientation notebook - they work!

PR #489, opened 2023-07-19
CI/CD - Adding installation checkers 

alex-rakowski:
This adds three different level of installation checks:
- quick makes sure any PR/Push to any branch, except main and dev, can be installed and imported on Python 3.10, on Ubuntu
- dev makes sure any PR/Push to dev can be installed and imported on python 3.9-3.11 on ubuntu
- main makes sure any PR/Push to main can be installed and imported  on python 3.9-3.11 on mac, ubuntu and windows, and checks but allows fail if it is compatible with 3.12

All have their ability to check against Apple silicon commented out to be ready when that image is available.


alex-rakowski:
We should probably update the `pypi_upload.yml` to be dependent on these checks.

bsavitzky:
These look great, thanks @alex-rakowski .  Let's wait until after the MM rush to review and accept these, they will be very helpful

alex-rakowski:
Sounds good, I misread an issue on the slack channel and thought this solved a pre M&M need

alex-rakowski:
Removed 3.9 checks

sezelt (approved):

PR #493, opened 2023-07-29
Updating whole pattern fitting for v 0.14.3

cophus:
Also added some plotting updates.


sezelt:
- [x] Improve distributed computation for better performance
- [x] Ability to link parameters between model objects
- [ ] Update all model objects to work with new linking framework
- [x] Add Moire model
- [ ] Add analytic gradients to Moire model

sezelt:
This PR includes big changes to how WPF works behind the scenes, with almost no user-facing differences. These changes greatly simplify the machinery for generating diffraction patterns inside the model, as well as allow for complicated linkages between model components, and improves fitting performance on many-core machines. 
* The mechanism for determining the correct index to read parameters and write derivative entries from the parameter vector is completely changed. The `Parameter` object now store their index into the global parameter vector, and the parent `WholePatternFit` object manages these indices. 
* This makes it easy for parameters of multiple models to be linked together, as they simply need to refer to the same `Parameter` object. The `WholePatternFit.link_parameters` method manages this process. 
* This also makes it simpler for models to enable/disable some fit parameters, because all indices are computed using this new machinery. Further, it makes index errors in general less likely because there are no more manual offsets. Some bugs of this nature were found and squashed, leading to faster convergence and better fits on some data we previously examined. 
* The global center coordinates are now simply `Parameter` objects tied to a `WPFModel`. Models that want to use the global center simply link to the `WholePatternFit.coordinate_model` model, which is always created by the parent and available under this special name. As a result, there are no longer separate functions for local versus global center coordinates in any of the models, removing a substantial amount of duplicated code. 
* Documentation of the WPF code is now much more complete
* The `SyntheticDiskMoire` model now allows fitting of Moire peaks arising from interference of two parent lattices. The `SyntheticDiskMoire` automatically finds the Moire unit cell and adds the corresponding peaks based on the parent lattices. 
* Parallel computation is now enabled by default, with multiple worker processes launched to operate on fits in parallel. This is accomplished with `mpire` as a wrapper around `multiprocessing`.  

image

sezelt:
I am still working on fixing the Moire unit cell determination but the rest of the code should be assumed finished and ready for review

smribet:
Overall this looks really great, thanks @sezelt  and @cophus! I'll wait to merge until the moire model is finalized.

smribet (approved):

PR #494, opened 2023-08-01
small changes to show

smribet:
This PR has some formatting and reorganizational changes to the show function to make it more readable, and an added option to plot the 2D FFT of an image.


cophus:
The show_fft option won't modify `ar` inplace right?

I noticed the:
        _ar = ar.copy()
line is below it.

smribet:
I don't think it modifies the array but just to be safe ...

sezelt (approved):

PR #499, opened 2023-08-19
reformat repo

smribet:
This PR reformats the repo. I think it will make it easier to compare and maintain code.

PR #502, opened 2023-08-23
Add a GitHub action that checks for `black` styling

sezelt:
Conditional on approving #499, it would make sense to enforce `black` styling on any updates to the repo. This should also prevent the situation where PRs contain a mix of new code and re-formatting of old code, which makes changes harder to comprehend.


.github/workflows/build-flake.yml
```diff
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a single version of Python
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: Python application
```
bsavitzky, line 4: THANK YOU

.github/workflows/build-flake.yml
```diff
@@ -1,7 +1,7 @@
...
+name: Check module can be imported
```
bsavitzky, line 4: THANK YOU

bsavitzky (approved):

PR #503, opened 2023-08-23
Listwrite patch

bsavitzky:
Patch to writing lists of data.  The primary changes are in `emdfile`, this PR only adds a unit test and updates `setup.py` to use the newest emdfile version, 0.0.12


sezelt (approved):
Thanks, @bsavitzky!

PR #505, opened 2023-08-29
nudging python version

alex-rakowski:
updating python version requirement to `>=3.10`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
     author_email='ben.savitzky@gmail.com',
     license='GNU GPLv3',
     keywords="STEM 4DSTEM",
-    python_requires='>=3.9,<3.12',
+    python_requires='>=3.10,<3.12',
     install_requires=[
         'numpy >= 1.19',
         'scipy >= 1.5.2',


sezelt (approved):

bsavitzky:
@sezelt @alex-rakowski Hey y'all - in the future, let's please discuss a version update like the basic Python version minimum before merging.  Thank you!

Lots of folks use versions of Python that are not as recent as 3.10.  (word on the street is @lerandc perfers 3.7 :p <3).

Is there new syntax someone wants to include in source or a compatibility issue requiring 3.10?  Or is there another motivation?

alex-rakowski:
`wpf.py line 153` and other lines have the new `|` operator for type hinting. Not sure how widely used this is outside that. I only nudged the version as 3.9 won't work, so I adjusted the setup to reflect that rather than having an opinion on dropping support for 3.9

bsavitzky:
Gotcha, thanks Alex.  It's true that `Union` is a lot more characters, and, this also avoids importing from `typing`.

Not sure I understood the last comment - in my mind requiring `>=3.10` in `setup.py` is dropping support for 3.9, as users will not now be able to install the package from a Python 3.9 installation.  Are you thinking of 'support' as meaning something else, or did I misunderstand something else?

alex-rakowski:
Im not sure if that's the only 3.10 feature we have, maybe we have some
match statements? I'm trying to find a linter that highlights any 3.9
incompatible syntax, to see the extent we're no longer compliant with 3.9.

I think it's possible to import annotations from futures and being able to
use | with 3.9. But I need to check this, I only half read the article.

I'd say if it's just one or two files, which only uses the `|` operator, we
should replace with union and roll back to 3.9, it doesn't seem like much
to give up for more python version flexibility.

I changed the setup.py file as a hot fix to be compatible with the codebase.

PR #507, opened 2023-08-29
adding codespace config

alex-rakowski:
This adds a default config setting to repo for using GitHub codespace.

Launches a editable installation from dev, in a miniconda (base) environment with some useful VS code settings and extensions. It's still possible to spin up a workspace not from this config, and otherwise has no affect on things.


bsavitzky (commented):
Cool extension.

PR #508, opened 2023-08-29
patches logscale vis bug

bsavitzky:
Fixes a bug in `show` which breaks when `scaling='log'` is passed and the array is all zeros.


cophus (approved):
Looks good to me

PR #510, opened 2023-08-29
Py39 fix

alex-rakowski:
adding back python 3.9 compatability, by using `from __future__ import annotations` in `wpf.py`, which has made it install with python 3.9 environment. I haven't run other tests, maybe @Steve could double-check?

### DIFF
--- a/.github/workflows/check_install_dev.yml
+++ b/.github/workflows/check_install_dev.yml
@@ -16,7 +16,7 @@ jobs:
                 allow_failure: [false]
                 runs-on: [ubuntu-latest]
                 architecture: [x86_64]
-                python-version: ["3.10", "3.11",]
+                python-version: ["3.9", "3.10", "3.11",]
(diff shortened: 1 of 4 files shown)


sezelt:
It seems pretty likely that before the usage of the new type hint syntax in `wpf.py` that we were Python 3.9 compatible. 

As far as I know, we are not using `match` statements anywhere at the moment, but I can see them being super useful in a few places. I think a lot of code where we base our control flow on the shape of nested tuples could be improved that way, for example.

alex-rakowski:
agreed, they'll be super useful. But until we use them, this PR will allow us to keep support for 3.9 with no rewrites as far as I can tell.

bsavitzky (approved):

bsavitzky:
It's definitely true that `match` could make some if/else clauses read a little more cleanly :)

PR #513, opened 2023-08-29
BraggVects append fix

bsavitzky:
Fixes a bug in which saving in "appendover" mode failed for Custom type objects, like BraggVectors.

Note to devs! @smribet @sezelt @alex-rakowski @cophus @gvarnavi:
This PR bumps the `emdfile` version, which is required for good behavior.  *Most likely* this will not cause any issues in your dev environments after merging/pulling on this code, however, if it does, re-run `pip install -e .` to update emdfile.

### DIFF
--- a/py4DSTEM/braggvectors/braggvectors.py
+++ b/py4DSTEM/braggvectors/braggvectors.py
@@ -285,6 +285,7 @@ class BraggVectors(Custom, BraggVectorMethods, Data):
         md["Qshape"] = self.Qshape
         self.metadata = md
         grp = Custom.to_h5(self, group)
+        return grp
 
     # read
 
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
         "gdown >= 4.7.1",
         "dask >= 2.3.0",
         "distributed >= 2.3.0",
-        "emdfile >= 0.0.12",
+        "emdfile >= 0.0.13",
         "mpire >= 2.7.1",
         "threadpoolctl >= 3.1.0",
     ],


alex-rakowski:
@bsavitzky py4DSTEM should always be using the latest version of `emdfile` right? If so I think we could setup an action to update `setup.py` to the latest version of `emdfile` every time there's a push to `emdfile/main`

alex-rakowski (approved):

bsavitzky:
Thanks for the suggestion @alex-rakowski - I think it's best to hold off on an action like that.  Each GH action requires maintenance, and is another fail point for an automation pipeline, and I think in a case like this automation really isn't needed.  Updating the version in setup.py was a trivial change/commit in a PR I had to make in to this repo anyway.  So I'd say its more complication than it's worth.

On the note of failure points in automation pipelines!  Something seems to have gone wrong with the read-the-docs build.  Remind me where we trigger a re-run of that process?  It's not with the other actions under Checks... :'''(

alex-rakowski:
It looks like there was an issue with build the docs website. I just manually ran it there, and it built with no issues, so I think it's good to go.

bsavitzky:
Cool, thanks Alex.

It'd be nice to be able to re-run this check the same way we re-run all the checks which are GH actions configured through .github/workflows/.yaml, but that has involved various hoops.  I modified the read-the-docs webhook so that it's supposed to run when you run workflows, instead of just on PR, but that didn't seem to do it.  Then the option to re-run the failed build through read-the-docs just apparently randomly appeared where I swear it wasn't this morning so that worked?  Maybe because of the webhook modification.....? That doesn't make sense to me, but, it worked! And now there are seven little green checkmarks and I am happy

PR #515, opened 2023-08-30
Virtual imaging shifting

smribet:
This PR allows for subpixel shifting of virtual detectors. 

@bsavitzky - I had a use case where I wanted to shift the detectors, but didn't want to use the centered geometry. I updated the defaults we had discussed previously a bit, so it'd be great if you could take a look.

### DIFF
--- a/py4DSTEM/datacube/virtualdiffraction.py
+++ b/py4DSTEM/datacube/virtualdiffraction.py
@@ -3,13 +3,12 @@
 #  * DataCubeVirtualDiffraction - methods inherited by DataCube for virt diffraction
 
 import numpy as np
-import dask.array as da
 from typing import Optional
 import inspect
 
 from emdfile import tqdmnd, Metadata
-from py4DSTEM.data import Calibration, DiffractionSlice, Data
-from py4DSTEM.visualize.show import show
+from py4DSTEM.data import DiffractionSlice, Data
+from py4DSTEM.preprocess import get_shifted_ar
 
 # Virtual diffraction container class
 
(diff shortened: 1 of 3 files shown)


py4DSTEM/datacube/virtualdiffraction.py
```diff
@@ -184,22 +183,33 @@ def get_virtual_diffraction(
             qx_shift = x0_mean - x0
             qy_shift = y0_mean - y0
 
-            # ...for integer shifts
-            if not subpixel:
+            if subpixel is False:
```
bsavitzky, line 186: why change `not subpixel` to `subpixel is False`? I think the former should be preferred

py4DSTEM/datacube/virtualdiffraction.py
```diff
@@ -210,29 +220,41 @@ def get_virtual_diffraction(
...
-                            virtual_diffraction += DP
-                        elif method == "max":
-                            virtual_diffraction = np.maximum(virtual_diffraction, DP)
-                    # normalize means
+                    # compute
                     if method == "mean":
-                        virtual_diffraction /= len(mask_indices[0])
+                        virtual_diffraction += DP
+                    elif method == "max":
+                        virtual_diffraction = np.maximum(virtual_diffraction, DP)
```
bsavitzky, line 227: missing an `else`

py4DSTEM/datacube/virtualdiffraction.py
```diff
@@ -243,21 +265,15 @@ def get_virtual_diffraction(
...
-                raise Exception("subpixel shifting has not been implemented yet!")
-                pass
+                    # compute
+                    w = mask[rx, ry]
+                    if method == "mean":
+                        virtual_diffraction += DP * w
+                    elif method == "max":
+                        virtual_diffraction = np.maximum(virtual_diffraction, DP * w)
+                if method == "mean":
+                    virtual_diffraction /= np.sum(mask)
```
bsavitzky, line 276: are these the behaviors we want for complex masks?

py4DSTEM/datacube/virtualimage.py
```diff
@@ -242,18 +245,27 @@ def _apply_mask_dask(self, mask):
                 self.calibration.get_origin_shift() is not None
             ), "origin need to be calibrated"
             qx_shift, qy_shift = self.calibration.get_origin_shift()
-            qx_shift = qx_shift.round().astype(int)
-            qy_shift = qy_shift.round().astype(int)
+            if subpixel is False:
```
bsavitzky, line 248: suggest `if not subpixel`

bsavitzky (approved):
Thanks @smribet !

I made a few minor comments, but nothing that I think should hold up incorporating these edits now so that they make it out with the 0.14.4 release.  The only comment that I would flag for you to check out is - in lines 272-276 of virtualdiffraction.py, are these the behaviors we want for complex masks?

PR #522, opened 2023-09-19
More strain less stress

smribet:
This PR is to update the strain branch by cleaning up the structure and adding more processing and plotting options.


smribet:
The linter is failing on the functions that are in py4DSTEM/process/latticevectors. 
@bsavitzky do you want to keep these functions or can I remove them? If we keep them, I just need to do a bit of restructuring.

PR #531, opened 2023-10-06
adding "sphinx_rtd_theme" to conf.py

alex-rakowski:
looks like this needed to be added to the extensions in conf.py 

https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 emdfile
+sphinx_rtd_theme
 # py4dstem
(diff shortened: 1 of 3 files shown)


sezelt:
Since this adds a dependency I think we should get a vibecheck from @bsavitzky. 

Alternatively, do you think it's possible to make this an optional dependency, like `py4dstem[docs]` and have the readthedocs builder install that option? That way ordinary installs don't require `sphinx_rtd_theme`.

sezelt (approved):
Having the requirement within the docs setup file seems like the best solution, since it doesn't interfere with the requirements for installing and running. 
Thanks

bsavitzky:
Thanks for working this out Steve and Alex! Agreed, this is the correct
solution :)

PR #532, opened 2023-10-09
update CUDA source file location in setup.py

sezelt:
This updates the location of the CUDA source files used in GPU disk detection so that they are packaged correctly.


bsavitzky (approved):
wooooo

PR #534, opened 2023-10-13
adding ASE compatibility to Crystal class

alex-rakowski:
Added two functions:
1) `from_ASE` creates Crystal from `ase.Atom` object
2) `from_generic_file` tries to create a Crystal using `ase.io.read` tested using a CIF file, but should add a lot of [file types](https://wiki.fysik.dtu.dk/ase/ase/io/io.html). This might need a better name.

### DIFF
--- a/py4DSTEM/process/diffraction/crystal.py
+++ b/py4DSTEM/process/diffraction/crystal.py
@@ -156,7 +156,7 @@ class Crystal:
 
         # Set occupancy attribute to None
         self.occupancy = None
-    
+
     def calculate_lattice(self):
         if not hasattr(self, "lat_real"):
             # calculate unit cell lattice vectors
@@ -426,20 +426,20 @@ class Crystal:
             numbers=atoms.numbers,
             cell=atoms.cell.array,
         )
-        
+
         # get occupancies
-        # ASE seems to have different ways of storing occupancies 
+        # ASE seems to have different ways of storing occupancies
         # if ASE object created from prismatic file
-        if 'occupancies' in atoms.arrays.keys():
+        if "occupancies" in atoms.arrays.keys():
             # np.array with length number of atoms
-            xtal.occupancy = atoms.arrays['occupancies'] 
+            xtal.occupancy = atoms.arrays["occupancies"]
             # if created from cif file
-        elif 'occupancy' in atoms.info.keys():
-            # python dict with occupancy per site 
-            xtal.occupancy = atoms.info['occupancy']
+        elif "occupancy" in atoms.info.keys():
+            # python dict with occupancy per site
+            xtal.occupancy = atoms.info["occupancy"]
         # TODO add in elif statement if other ways appear
         else:
-            print(Warning("Could not find occupancies of crystal")) 
+            print(Warning("Could not find occupancies of crystal"))
         return xtal
 
     def from_prismatic(filepath):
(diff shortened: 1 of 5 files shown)


alex-rakowski:
After a little playing around I'm not sure how useful the generic file reader is, I forgot that standard xyz files don't contain cell params. I think the majority of the other supported file types are MD related. So I think I'll remove the generic function and add in `from_prismatic` instead.

alex-rakowski:
Updated to have `from_Ase` and `from_prismatic`. I commented out the `from_generic_file` in case people wanted it back in, I can delete if preferred. Otherwise I think it should be good to go.

alex-rakowski:
Updated to have `from_Ase` and `from_prismatic`. I commented out the `from_generic_file` in case people wanted it back in, I can delete if preferred. Otherwise I think it should be good to go.

bsavitzky:
Thanks @alex-rakowski , this is a great idea!  Two questions:

(1) dependency: right now you import `ase` in a `try` statement and have not modified the setup file.  I definitely appreciate that this structure won't break the code for anyone.  All that said - 
- This might be worth just adding as a dependency.  ase is a long-lived and well supported module.
- If we don't do this, we can add it as an optional dependency. I don't really love adding tons of optional dependency options, but I like it better than having some secret pseudo-dependencies where you can use certain methods iff you've installed the relevant packages yourself.
What do you think?

(2) tests: our test suite is still far from robust...but its got better coverage in IO than anywhere else, and I'd love to ideally not reduce that coverage.  The pipeline for testing file read stuff should be pretty easy to add (a) new file(s) + test(s) to.  If you drop some new test files in our usual sample file drive and modify the `file_ids` and `collection_ids` dicts in `io/google_drive_downloader.py` to point to them you can make test files accessible to others (files should be SMALL, but shouldn't be an issue in this case).  You could then add tests in `test/test_nonnative_io` using the patterns in the other two files in that directory.

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -14,12 +14,19 @@
...
 try:
     from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
     from pymatgen.core.structure import Structure
 except ImportError:
     pass
 
+try:
+    from ase.io import read
+except ImportError:
+    pass
```
sezelt, line 28: For an optional dependency that is not needed at the top level of the file, I think a better pattern is to move the import within the function that requires it. That way any import errors get raised at the appropriate moment.

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -400,6 +410,149 @@ def from_pymatgen_structure(
...
+            filepath (str|Pathlib.Path): path to the file
+        """
+
+        # check if ase is installed
+        if find_spec("ase") is None:
+            raise ImportWarning(
+                "Could not import ASE, please install, restart and try again"
+            )
+        else:
+            from ase.io import read
```
sezelt, line 498: Is all this extra logic necessary? If `ase` is missing it will raise a ModuleNotFoundError

py4DSTEM/process/diffraction/crystal.py
alex-rakowski: I'm going to delete as its now a dependency

py4DSTEM/process/diffraction/crystal.py
alex-rakowski: I'm going to delete as its now a dependency

alex-rakowski:
@sezelt  Are you planning on creating a new PR for your CIF occupancy, adding into this one, or do you want to send me the code snippet and I can add it?

alex-rakowski:
closing as this will be wrapped in to #538

PR #547, opened 2023-10-27
Quick fix for correlation array, scalebar length

cophus:
Added new option to crystal.orientation_plan()
calculate_correlation_array = True,
Setting this to false skips calculating the correlation array.

Additionally a user pointed out that scalebars were being drawn slightly too long, so I added a quick fix for that too.


alex-rakowski (approved):

alex-rakowski:
I added an assert statement in `match_single_pattern` to check `orientation_plan` was run with `calculate_correlation_array = True` before running the rest of the function.

PR #552, opened 2023-11-01
Add robust fitting back to `fit_origin`

### LINKED ISSUE #548: Several arguments to `BraggVectors.fit_origin` are ignored
**Describe the bug**
In `BraggVectorMethods.fit_origin` the arguments `fitfunction`, `robust`, `robust_steps`, and `robust_thresh` are completely ignored. 

**To Reproduce**
See below code, the arguments are not used
https://github.com/py4dstem/py4DSTEM/blob/9a2675b37c56a6e6faa0657d4aca84b77f5b4a35/py4DSTEM/braggvectors/braggvector_methods.py#L511-L562

**Expected behavior**
I expect these to do things

**py4DSTEM version** 
0.14.8
**Python version**
3.11.0
**Operating system** 
Ubuntu

sezelt:
Fixes #548

### DIFF
--- a/py4DSTEM/braggvectors/braggvector_methods.py
+++ b/py4DSTEM/braggvectors/braggvector_methods.py
@@ -552,14 +552,19 @@ class BraggVectorMethods:
         from py4DSTEM.process.calibration import fit_origin
 
         if mask_check_data is True:
-            # TODO - replace this bad hack for the mask for the origin fit
-            mask = np.logical_not(q_meas[0] == 0)
-            qx0_fit, qy0_fit, qx0_residuals, qy0_residuals = fit_origin(
-                tuple(q_meas),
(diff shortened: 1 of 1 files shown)


bsavitzky (approved):
Looks good, thanks @sezelt.  Not sure how things wound up in this strange state.

There will be conflicts with the edits I made in my GTO branch to add the masking functionality correctly back in to this method, but I'll just pull from dev and handle them locally before PRing those changes.  I'll likely modify the arguments at that point as well (`mask_check_data` is nutso), so small modifications to user-side code may be necessary then, but these should be small things.

Lmk if you still need the masking mods given this PR.  Since I'll need to handle conflicts before I can push that stuff, I can try to get to that sooner if you need it or a bit later if not.

PR #554, opened 2023-11-04
I've got a repeat joke, baby, and I'll import your namespace

alex-rakowski:
@bsavitzky PR to make the namespace changes we discussed.

As implemented the namespace is quite cluttered, which makes autocomplete less useful. I'm wondering if it's worth pushing everything down one level. e.g. `py4DSTEM.process.StrainMap` (rather than `py4DSTEM.StrainMap` and the existing `py4DSTEM.process.strain.StrainMap`) and keeping the top level to `py4DSTEM.read/save/import_file/show/show_complex/check_config ` and modules `io/process/data etc.`


bsavitzky (commented):
Thanks @alex-rakowski ! Looks good overall. A few things:

1. Not sure if we want all the ptycho reconstruction classes in the main namespace. Wondering if having them down a level might be better, e.g. py4DSTEM.phase.XXX, ok with either this or that - curious what @gvarnavi and @smribet think
2. That said, I don't think we should be pushing everything down a level like this - most classes should be in the main namespace! I.e. as you've done it here :)
3. There appear to now be merge conflicts - are you able to resolve those?

py4DSTEM/__init__.py
```diff
@@ -52,18 +52,74 @@
...
-# TODO - ptycho
-# TODO - others
-
-# TODO - where
-from py4DSTEM.process import (
-    PolarDatacube,
-)
-
+# TODO WPF
+from py4DSTEM.process import wholepatternfit
```
bsavitzky, line 106: @sezelt the wholepatternfit namespace is just two `import *` statements - this would be great to tidy at some point :p

bsavitzky (commented):
Looks good, once the extra commented lines are removed this can be merged!

bsavitzky (approved):

PR #557, opened 2023-11-08
I knew you were (h,k,l) when you (g1,g2)'d in

alex-rakowski:
- Added functionality to take g1 g2 indexed and convert to hkl if g1_hkl,g2_hkl are known, 
- changed Strainmap to have bragg_vectors_indexed indexed vectors as `g1_ind`, and `g2_ind` instead of `h`, and `k`
- updated Bragg index overlay func to accept pointillist with fields `g1_ind`, `g2_ind`, and `h`,`k`,`l`


alex-rakowski:
need to check that the [below](https://github.com/alex-rakowski/py4DSTEM/blob/283f8eab0b61eeeaffab2b143e53e99761ba1dd3/py4DSTEM/process/strain/strain.py#L406-L412) is correct:
```
        g1_hkl = np.array(g1_hkl)
        g2_hkl = np.array(g2_hkl)

        # TODO check this is correct
        # calculate the direct beam
        direct_beam = np.cross(g1_hkl, g2_hkl)
        direct_beam = direct_beam / np.gcd.reduce(direct_beam)
 ```

**Comment by alex-rakowski** [comment]
added:

```python 
# make h,k,l data available like qx,qy,I,data, Handles errors if h,k,l not set.
braggpeaks.cal[0,0].h/.k/.l

# adds call to py4DSTEM.process.diffraction.plot_diffraction_pattern from 
# BraggVectors
braggpeaks.plot(index=(0,0)) 
# and BraggVect objects.
braggpeaks.cal[0,0].plot()
```

bsavitzky (approved):

PR #558, opened 2023-11-08
'Cause PEP-8 never goes out of style

alex-rakowski:
pep 8 - Comparisons to singletons compliance
```
# goes from 
if arg == True:
    pass
if False == arg:
    pass
# to
if arg is True:
    pass
if arg is False:
    pass
```

### DIFF
--- a/py4DSTEM/process/diffraction/crystal_ACOM.py
+++ b/py4DSTEM/process/diffraction/crystal_ACOM.py
@@ -1840,7 +1840,9 @@ def cluster_grains(
             xr = np.clip(x + np.arange(-1, 2, dtype="int"), 0, sig.shape[0] - 1)
             yr = np.clip(y + np.arange(-1, 2, dtype="int"), 0, sig.shape[1] - 1)
             inds_cand = inds_all[xr[:, None], yr[None], :].ravel()
-            inds_cand = np.delete(inds_cand, mark.ravel()[inds_cand] == False) # noqa: E712,E501
+            inds_cand = np.delete(
+                inds_cand, mark.ravel()[inds_cand] == False  # noqa: E712
+            )
 
             if inds_cand.size == 0:
                 grow = False
@@ -1893,7 +1895,7 @@ def cluster_grains(
 
                 inds_grain = np.append(inds_grain, inds_cand[keep])
                 inds_cand = np.unique(
-                    np.delete(inds_new, mark.ravel()[inds_new] == False) #noqa: E712
+                    np.delete(inds_new, mark.ravel()[inds_new] == False)  # noqa: E712
                 )
 
                 if inds_cand.size == 0:
--- a/py4DSTEM/process/diffraction/crystal_ACOM.py
+++ b/py4DSTEM/process/diffraction/crystal_ACOM.py
@@ -1840,7 +1840,7 @@ def cluster_grains(
             xr = np.clip(x + np.arange(-1, 2, dtype="int"), 0, sig.shape[0] - 1)
             yr = np.clip(y + np.arange(-1, 2, dtype="int"), 0, sig.shape[1] - 1)
             inds_cand = inds_all[xr[:, None], yr[None], :].ravel()
-            inds_cand = np.delete(inds_cand, mark.ravel()[inds_cand] is False)
+            inds_cand = np.delete(inds_cand, mark.ravel()[inds_cand] == False) # noqa: E712,E501
 
             if inds_cand.size == 0:
                 grow = False
@@ -1893,7 +1893,7 @@ def cluster_grains(
 
                 inds_grain = np.append(inds_grain, inds_cand[keep])
                 inds_cand = np.unique(
-                    np.delete(inds_new, mark.ravel()[inds_new] is False)
+                    np.delete(inds_new, mark.ravel()[inds_new] == False) #noqa: E712
                 )
 
                 if inds_cand.size == 0:
--- a/py4DSTEM/process/diffraction/crystal_ACOM.py
+++ b/py4DSTEM/process/diffraction/crystal_ACOM.py
@@ -798,12 +798,12 @@ def match_orientations(
     )
 
     # check cal state
-    if bragg_peaks_array.calstate["ellipse"] == False:
+    if bragg_peaks_array.calstate["ellipse"] is False:
         ellipse = False
         warn("Warning: bragg peaks not elliptically calibrated")
     else:
         ellipse = True
-    if bragg_peaks_array.calstate["rotate"] == False:
+    if bragg_peaks_array.calstate["rotate"] is False:
         rotate = False
         warn("bragg peaks not rotationally calibrated")
     else:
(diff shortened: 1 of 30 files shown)


py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -1840,7 +1840,7 @@ def cluster_grains(
             xr = np.clip(x + np.arange(-1, 2, dtype="int"), 0, sig.shape[0] - 1)
             yr = np.clip(y + np.arange(-1, 2, dtype="int"), 0, sig.shape[1] - 1)
             inds_cand = inds_all[xr[:, None], yr[None], :].ravel()
-            inds_cand = np.delete(inds_cand, mark.ravel()[inds_cand] == False)
+            inds_cand = np.delete(inds_cand, mark.ravel()[inds_cand] is False)
```
sezelt, line 1843: This is different behavior

py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -1893,7 +1893,7 @@ def cluster_grains(
 
                 inds_grain = np.append(inds_grain, inds_cand[keep])
                 inds_cand = np.unique(
-                    np.delete(inds_new, mark.ravel()[inds_new] == False)
+                    np.delete(inds_new, mark.ravel()[inds_new] is False)
```
sezelt, line 1896: Same

py4DSTEM/process/fit/fit.py
```diff
@@ -107,7 +107,7 @@ def fit_2D(
                 fit_mean_square_error
                 > np.mean(fit_mean_square_error) * robust_thresh**2
             )
-            mask[_mask] == False
+            mask[_mask] is False
```
sezelt, line 110: I'm not sure what the intended behavior of the original code was here. It looks like this should have been an assignment.

py4DSTEM/process/polar/polar_peaks.py
```diff
@@ -167,7 +167,7 @@ def find_peaks_single_pattern(
     if remove_masked_peaks:
         peaks = np.delete(
             peaks,
-            mask_bool[peaks[:, 0], peaks[:, 1]] == False,
+            mask_bool[peaks[:, 0], peaks[:, 1]] is False,
```
sezelt, line 170: Different behavior

py4DSTEM/process/strain/latticevectors.py
```diff
@@ -182,7 +182,7 @@ def fit_lattice_vectors(braggpeaks, x0=0, y0=0, minNumPeaks=5):
 
     # Remove unindexed peaks
     if "index_mask" in braggpeaks.dtype.names:
-        deletemask = braggpeaks.data["index_mask"] == False
+        deletemask = braggpeaks.data["index_mask"] is False
```
sezelt, line 185: Same

py4DSTEM/process/strain/strain.py
```diff
@@ -791,17 +791,17 @@ def show_strain(
 
         # Get images
         e_xx = np.ma.array(
-            self.get_slice("exx").data, mask=self.get_slice("mask").data == False
+            self.get_slice("exx").data, mask=self.get_slice("mask").data is False
```
sezelt, line 794: Same

py4DSTEM/process/strain/strain.py
```diff
@@ -791,17 +791,17 @@ def show_strain(
...
         )
         e_yy = np.ma.array(
-            self.get_slice("eyy").data, mask=self.get_slice("mask").data == False
+            self.get_slice("eyy").data, mask=self.get_slice("mask").data is False
```
sezelt, line 797: Same

py4DSTEM/process/strain/strain.py
```diff
@@ -791,17 +791,17 @@ def show_strain(
...
         )
         e_xy = np.ma.array(
-            self.get_slice("exy").data, mask=self.get_slice("mask").data == False
+            self.get_slice("exy").data, mask=self.get_slice("mask").data is False
```
sezelt, line 800: Same

py4DSTEM/process/strain/strain.py
```diff
@@ -791,17 +791,17 @@ def show_strain(
...
         )
         theta = np.ma.array(
             self.get_slice("theta").data,
-            mask=self.get_slice("mask").data == False,
+            mask=self.get_slice("mask").data is False,
```
sezelt, line 804: Same

sezelt (changes requested):
For builtin Python booleans, testing `== False` and `is False` have the same behavior, *but for other types this is not the case.* In particular for `numpy` arrays, these do entirely different things: `is` compares the identity of the array object itself, while `==` is an elementwise value comparison. For example:
```python
In [1]: import numpy as np

In [2]: ar = np.random.rand(3,3)

In [3]: ar
Out[3]:
array([[0.12607551, 0.56005212, 0.42254926],
       [0.48698435, 0.5986023 , 0.75294198],
       [0.82027795, 0.58230241, 0.92049683]])

In [4]: b = ar > 0.5

In [5]: b
Out[5]:
array([[False,  True, False],
       [False,  True,  True],
       [ True,  True,  True]])

In [6]: b is False
Out[6]: False

In [7]: b == False
Out[7]:
array([[ True, False,  True],
       [ True, False, False],
       [False, False, False]])
```

There is also one place where I can't tell the original meaning of the code, but from context it looks like it should actually be an assignment (`=`) rather than an equality test (see review comment).

py4DSTEM/process/fit/fit.py
alex-rakowski: changing to

alex-rakowski:
Thanks @sezelt

sezelt (approved):
Thanks for the changes. I think we are just waiting on clarification of that one line whose original meaning was unclear

py4DSTEM/process/fit/fit.py
sezelt: It might blame to @smribet because of the big reformatting PR. I suspect @cophus or @bsavitzky wrote the code here

sezelt (changes requested):
Just needs clarification on the line in fit.py

py4DSTEM/process/fit/fit.py
bsavitzky: Pretty sure this was me.  Why are we changing every boolean `==` to `is` now?  Does this PR mean I have to read PEP-8?  I feel like this is not universally desirable behavior...

bsavitzky:
So this is just changing every boolean `==` to an `is`?  Not sure I'm into it :p

sezelt:
From PEP-8:
> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

See also this Stackoverflow:
https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true

This thread actually shows an interesting example where a numpy function that returns a scalar boolean does *not* return a Python singleton, and thus would behave differently under this change.

alex-rakowski:
It changes `==` into `is` for comparisons to singletons, which is the [pep-8 style](https://peps.python.org/pep-0008/#programming-recommendations).

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.
> Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

@sezelt caught the examples I incorrectly changed, which were used in numpy masks.

py4DSTEM/process/fit/fit.py
bsavitzky: Ok I reviewed and this (a) was Colin and (b) def should be an assignment :p

alex-rakowski: changed to assignment

alex-rakowski:
@sezelt @bsavitzky Changed to assignment should be good to go

bsavitzky (approved):

PR #563, opened 2023-11-13
Remove linter action redundancies

bsavitzky:
Remove redundancies and fix opaque naming conventions in linter actions


mlhenderson:
Expect a bunch of linter errors after the config is resolved that will need to be sorted out in this PR.

alex-rakowski:
@bsavitzky  these are [my suggested changes](https://github.com/bsavitzky/py4DSTEM/pull/3) for the linter

.flake8
```diff
@@ -0,0 +1,17 @@
...
+    E201,
+    E202,
+    E203,
+    E204,
+    E231,
+    E265,
+    E266,
+    E303,
+    E402,
+    E501,
```
alex-rakowski, line 15: I think we should add, locally collects files differently to the GA 
to make it ignore `__init__.py` and `setup.py` files and to stop it searching docs, when run locally, we can also add to tests it to ignore this as well
```
exclude =
    *__init__.py
    *setup.py
    docs
 ```

PR #569, opened 2023-11-15
righting my wrongs 

alex-rakowski:
Hotfix - deletemask[i] == False

### DIFF
--- a/py4DSTEM/braggvectors/threshold.py
+++ b/py4DSTEM/braggvectors/threshold.py
@@ -52,7 +52,7 @@ def threshold_Braggpeaks(
             r2 = minPeakSpacing**2
             deletemask = np.zeros(pointlist.length, dtype=bool)
             for i in range(pointlist.length):
-                if deletemask[i] is False:
+                if deletemask[i] == False:  # noqa: E712
                     tooClose = (
                         (pointlist.data["qx"] - pointlist.data["qx"][i]) ** 2
                         + (pointlist.data["qy"] - pointlist.data["qy"][i]) ** 2
@@ -160,7 +160,7 @@ def universal_threshold(
             r2 = minPeakSpacing**2
             deletemask = np.zeros(pointlist.length, dtype=bool)
             for i in range(pointlist.length):
-                if deletemask[i] is False:
+                if deletemask[i] == False:  # noqa: E712
                     tooClose = (
                         (pointlist.data["qx"] - pointlist.data["qx"][i]) ** 2
                         + (pointlist.data["qy"] - pointlist.data["qy"][i]) ** 2
--- a/py4DSTEM/braggvectors/threshold.py
+++ b/py4DSTEM/braggvectors/threshold.py
@@ -1,7 +1,7 @@
 # Bragg peaks thresholding fns
 
 import numpy as np
-
+import numpy.typing as npt
 from emdfile import tqdmnd, PointListArray
 
 
@@ -158,7 +158,7 @@ def universal_threshold(
         # Remove peaks that are too close together
         if maxNumPeaks is not False:
             r2 = minPeakSpacing**2
-            deletemask = np.zeros(pointlist.length, dtype=bool)
+            deletemask: npt.NDArray = np.zeros(pointlist.length, dtype=bool)
             for i in range(pointlist.length):
                 if deletemask[i] == False:  # noqa: E712
                     tooClose = (
--- a/py4DSTEM/braggvectors/threshold.py
+++ b/py4DSTEM/braggvectors/threshold.py
@@ -1,7 +1,6 @@
 # Bragg peaks thresholding fns
 
 import numpy as np
-import numpy.typing as npt
 from emdfile import tqdmnd, PointListArray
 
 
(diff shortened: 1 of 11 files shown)


bsavitzky (commented):
Looks fine - there appears to be a merge conflict now, however.  Guessing this may have to do with the small PR Steph made to fix the issue that she ran into - perhaps you also duplicated that fix here but slightly differently.  Could you resolve the merge conflicts @alex-rakowski ?

alex-rakowski:
@bsavitzky resloved - it was the same logic as Steph's change, but I've added a `# noqa` statement to block the linter from raising an error on this next time

bsavitzky:
Gotcha, that was my guess :p thank you! I'll review and merge when I'm cak
at a computer

py4DSTEM/braggvectors/threshold.py
```diff
@@ -158,9 +158,9 @@ def universal_threshold(
         # Remove peaks that are too close together
         if maxNumPeaks is not False:
             r2 = minPeakSpacing**2
-            deletemask = np.zeros(pointlist.length, dtype=bool)
+            deletemask: npt.NDArray = np.zeros(pointlist.length, dtype=bool)
```
sezelt, line 161: Why are we adding a type hint here? I didn't even know you could do inline type hints like this!

py4DSTEM/process/calibration/ellipse.py
```diff
@@ -63,7 +63,7 @@ def fit_ellipse_1D(ar, center=None, fitradii=None, mask=None):
     rr = np.sqrt((xx - x0) ** 2 + (yy - y0) ** 2)
     _mask = (rr > ri) * (rr <= ro)
     if mask is not None:
-        _mask *= mask is False
+        _mask *= mask == False  # noqa: E712
```
sezelt, line 66: This should be converted to a binary logical operation. I think this is effectively doing `logical_and`?

sezelt (changes requested):
I've gone through the changes made by #558 and tried to make sure that any remaining instances where values that may not be a Python singleton are being compared correctly. I think the following places are still likely errors (though some of these may have been fixed by the series of hotfixes that we've made following #558). 
* read_K2.py:400
* radialbkgrd.py:83
* origin.py:157,165

Still, I can't say I'm certain that we are always passing the True/False singletons in all of the other cases, so I am somewhat inclined to simply revert #558, especially given the very subtle nature of the errors that have cropped up so far...

py4DSTEM/braggvectors/threshold.py
alex-rakowski: I didn't mean to include this, while valid code, I just wanted to see the state of type inferencing on linters but decided it wasn't worth it. I'll remove it and the import of npt

py4DSTEM/process/calibration/ellipse.py
alex-rakowski: I was just changing this back to the state before #558, and added a comment to not raise an error in linters.

alex-rakowski:
> * read_K2.py:400 -- 

I thought it was set to [singleton](https://github.com/py4dstem/py4DSTEM/blob/40ae3cbe6c9ed5ccc826edd1c378a419736ebd7a/py4DSTEM/io/filereaders/read_K2.py#L398) but can revert if preferred. 

> * radialbkgrd.py:83 --

 [.all()](https://github.com/py4dstem/py4DSTEM/blob/40ae3cbe6c9ed5ccc826edd1c378a419736ebd7a/py4DSTEM/preprocess/radialbkgrd.py#L83C15-L83C51)  converts to singleton But I can change back to  == if preferred. 

> * origin.py:157,165 -- 

has been [fixed](https://github.com/py4dstem/py4DSTEM/blob/40ae3cbe6c9ed5ccc826edd1c378a419736ebd7a/py4DSTEM/process/calibration/origin.py#L157)


If reverting #558  is preferred then that also works, but I think a series of hot patches and this has covered it. But I have no preference.

sezelt:
`.all()` does *not* return a Python singleton
```python
In [1]: import numpy as np

In [2]: ar = np.ones((3,3),dtype=np.bool_)

In [3]: ar.all() == True
Out[3]: True

In [4]: ar.all() is True
Out[4]: False

In [5]: type(ar.all())
Out[5]: numpy.bool_

In [6]: type(True)
Out[6]: bool
```

alex-rakowski:
> `.all()` does _not_ return a Python singleton
> 
> ```python
> In [1]: import numpy as np
> 
> In [2]: ar = np.ones((3,3),dtype=np.bool_)
> 
> In [3]: ar.all() == True
> Out[3]: True
> 
> In [4]: ar.all() is True
> Out[4]: False
> 
> In [5]: type(ar.all())
> Out[5]: numpy.bool_
> 
> In [6]: type(True)
> Out[6]: bool
> ```
Thanks, I'll fix this and go back over the commits to see if I assumed this  or `any()` elsewhere

alex-rakowski:
@sezelt  I've gone through the changes in #558 and put everything that isn't typed as a bool back to `==`, so I think this should be good to go. The final remaining part:
> This should be converted to a binary logical operation. I think this is effectively doing `logical_and`? in [py4DSTEM/process/calibration/ellipse.py](https://github.com/py4dstem/py4DSTEM/pull/569/files/dbaf3c68184296d36fd54b279891fe15a232baca#diff-1e8c4e76370bde19e5e267f9ef39f8458a8d6db3bcf3eb145cdd97a802e5693a), 

I just moved this back to pre #558 It may be `logical_and` but maybe that should be a separate PR?

sezelt (approved):
Thanks, this looks good as far as I can tell

PR #560, opened 2023-11-10
Clearing up some bare exceptions

alex-rakowski:
Getting rid of some of the bare exceptions. Adding the correct Error if I could work it out otherwise using `Exception`.

PR #571, opened 2023-11-16
F811, F523/F524, F841, F741 - Corrections in crystal.py

alex-rakowski:
F811, F523/F524, F841, F741 - Corrections


alex-rakowski:
added the changes to `linter.yaml`, `.flake8` and removed `build-flake.yaml` as the GA is in a strange state in dev at the moment, so the test would fail.

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -660,7 +649,8 @@ def generate_diffraction_pattern(
             self.setup_diffraction(300e3)
 
         # Tolerance for angular tests
-        tol = 1e-6
+        # TODO should this be used somewhere ?
```
sezelt, line 652: This does seem to be unused, we can delete it

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -816,7 +806,8 @@ def generate_ring_pattern(
 
         # check accelerating voltage
         if hasattr(self, "accel_voltage"):
-            accelerating_voltage = self.accel_voltage
+            # TODO should this be used somewhere ?
+            accelerating_voltage = self.accel_voltage  # noqa: F841
```
sezelt, line 810: This is also unused, can be deleted.

sezelt (changes requested):
The changes look good, you can delete those two unused variables in `crystal.py` and then I think this is ready to go.

sezelt:
This does seem to overlap with #563 a bit, so maybe they should be combined? Probably the only potential conflict is if they make different changes to the flake8 file

py4DSTEM/process/diffraction/crystal.py
alex-rakowski: removed

py4DSTEM/process/diffraction/crystal.py
alex-rakowski: removed

py4DSTEM/process/diffraction/crystal.py
```diff
@@ -815,9 +801,7 @@ def generate_ring_pattern(
             )
 
         # check accelerating voltage
-        if hasattr(self, "accel_voltage"):
-            accelerating_voltage = self.accel_voltage
-        else:
+        if not hasattr(self, "accel_voltage"):
```
alex-rakowski, line 804: @sezelt deleted the var, but this should preserve the functionality

sezelt (approved):

sezelt:
I'm approving this since all the changes it makes are acceptable, but I'll hold off on merging so that we can decide how to reconcile with #563

bsavitzky (changes requested):
See comment in-thread

bsavitzky:
Hey @alex-rakowski - 

This PR is redundant / in conflict with PR#563.  All file lines changed here other than those in crystal.py have already been modified, often quite differently, in that PR.  Can we sync up?

Ultimately this should be a single PR, otherwise we're just making more work for ourselves!

sezelt:
I've merged this with #560

PR #579, opened 2023-11-22
Fixing Google Colab Issues

alex-rakowski:
This PR changes the except behaviour for `Cupy` and `pylops` so that in envrionments which have cupy installed incorrectly py4DSTEM can still import correctly. Its a patch fix, and it might be worth thinking how the Exceptions are handled and if they should raise warnings about the state of cupy imported, as it may currently just convert cupy to numpy somewhat silently. 

`pylops` was throwing an error as it determines the cupy envrionment by:

```python
cupy_enabled = (
    util.find_spec("cupy") is not None and int(os.getenv("CUPY_PYLOPS", 1)) == 1
)
```
which when `cupy` is installed incorrectly can still return True.

I found this using:

```python
import traceback

try:
  import py4DSTEM
except Exception:
    print(traceback.format_exc())
```

### DIFF
--- a/py4DSTEM/braggvectors/diskdetection_cuda.py
+++ b/py4DSTEM/braggvectors/diskdetection_cuda.py
@@ -4,8 +4,9 @@ Functions for finding Braggdisks using cupy
 """
 
 import numpy as np
-import cupy as cp
+
 # TODO Wanted to double check these are good unprotected
+import cupy as cp
 from cupyx.scipy.ndimage import gaussian_filter
 import cupyx.scipy.fft as cufft
 from time import time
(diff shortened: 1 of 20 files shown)


py4DSTEM/braggvectors/diskdetection_cuda.py
```diff
@@ -4,6 +4,8 @@
 """
 
 import numpy as np
+
+# TODO Wanted to double check these are good unprotected
```
sezelt, line 8: This file is only imported when CUDA is selected, so it's safe

py4DSTEM/braggvectors/kernels.py
```diff
@@ -1,3 +1,4 @@
+# TODO check these are okay to leave unprotected
```
sezelt, line 1: This is also imported only when using CUDA

sezelt (commented):
I guess it turns out that lots of people have deliberately misconfigured `cupy` in their environments, so this seems to be the only way to go.

bsavitzky (approved):
Thanks @alex-rakowski !  To the rescue!

PR #581, opened 2023-11-22
MRS hot fixes

smribet:
This PR has a refactor of the strain plotting functions. The strain notebooks and WS2 ACOM example work here.
[tutorials.zip](https://github.com/py4dstem/py4DSTEM/files/13445047/tutorials.zip)


alex-rakowski (approved):
Looks good. Only comment is we now have `exx` and `e_xx` etc. used in different places, and should we standardise to one across everything at somepoint

alex-rakowski (approved):
Looks good. Only comment is we now have `exx` and `e_xx` etc. used in different places, and should we standardise to one across everything at somepoint

PR #586, opened 2023-12-12
Adding coefficient calculations to orientation correlations

cophus:
We can now set:
py4DSTEM.process.diffraction.plot_orientation_correlation(
    orient_corr,
    calculate_coefs = True,
    ...
)
Which will calculate and optionally draw:
-The user specified % probability for radial and annular lengths
-The slope of the zero intercept in degrees / pixel_units


cophus:
These are the new options / outputs:
image (4)

And this is what the plot overlays look like:
image (5)

sezelt (approved):
Looks good to me

PR #592, opened 2023-12-31
Update corr_kernel_size documentation

alex-rakowski:
Quick patch to correct the units, and provide more information as to what its parameter does (captured from slack).


sezelt (approved):
ok

PR #595, opened 2024-01-03
adding dependabot functionality 

alex-rakowski:
This adds dependabot functionality to check and create PRs for updates in our GitHub actions dependencies. Similar to the vulnerability one was raised in #593 . 

We could extend this to check pip packages, but I've left this out as it is a more breaking change. 

```
  - package-ecosystem: "pip" # See documentation for possible values
    directory: "/" # Location of package manifests
    schedule:
      interval: "weekly"
      day: "sunday"
      time: "23:00"
    ignore:
      - dependency-name: "tensorflow"
      - dependency-name: "tensorflow-addons"
      
 ```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "23:00"


sezelt:
What would actually break if we were to enable checking of pip dependencies?

alex-rakowski:
The ones I'd be worried about are, tensorflow/-addons and mp-api.


I haven't tested the latest version of mp-api, we might be able to loosen
it to just exclude some versions. Or kick the problem down the road and add
it to the ignore packages.

sezelt:
I guess I'm more asking, what will it actually do in that case? Will it open random PRs? Will it make tests fail? In what way will things actually be "broken?"

alex-rakowski:
It opens a pr per dependency to upgrade, I think we can set it to group all
the upgrades into a single pr, but this isn't the default.

I'm not 100% on the behaviour if you decline a pr, and if it will open a pr
to bump the version to a previously rejected one the next time it runs or
not.  E.g. it wants to bump to x to 1.3, we decline, next Sunday it might
suggest X to 1.3, or it might wait to bump x to 1.31.

None of our actions test functionality, so unless they break the install or
import, we'd wouldn't know if they were breaking upgrades.

sezelt:
OK in that case, I suppose we don't want to turn on pip packages. For github actions, will this do anything different than what we already have? There seems to be some default level of checking for actions vulnerabilities, since we got that other PR. Is there some severity setting that might be different?

alex-rakowski:
The pr we got came from a setting that looks only for vulnerabilities.

These prs would keep things up to date,  which would help avoid
depreciation issues, get speed/feature bumps etc. I've had it running on my
fork for a little while and the updates are fairly infrequent and haven't
broken anything so far.

bsavitzky (approved):
Cool!  Looks good :)

PR #596, opened 2024-01-03
Update check_config workflow

alex-rakowski:
This pull request updates the check_config workflow. Taken from #560. 

This PR does the following:

Updates config_checker so that it does scrapes the required modules and optional installs from setup.py and adds generic version checking for installed versions of packages. New behavior is if we add a dependency or add a new optional set of dependencies e.g. ~"dev" : ["pytest", "black", "flake8"], it will no automatically pick up on "dev" install optional and test whether all the dependencies are met, and report the state i.e. all met or not. Previously, we had to manually update by adding a dependency or optional dependencies set to a list or dict.


sezelt (approved):
This looks like a good improvement and should improve maintainability

PR #597, opened 2024-01-08
You get a Mixin, you get a Mixin, every ptycho class gets a Mixin

gvarnavi:
Draft PR for long-awaited `phase_contrast` refactor using Mixins to de-duplicate code.  
Also includes various bugfixes, improvements, and new features.  
Will update description soon with more information and attach a notebook to help with reviewing.

PR #599, opened 2024-01-22
Release memory after Bayesian optimization

sezelt:
Previously the BO code retained a reference to the ptycho reconstruction, which in some cases would include a copy of the dataset. This ensures that these resources are released. Additionally it puts `pbar.close()` into a `try/finally` clause so the progressbar is always terminated correctly.

### DIFF
--- a/py4DSTEM/braggvectors/diskdetection_cuda.py
+++ b/py4DSTEM/braggvectors/diskdetection_cuda.py
@@ -156,9 +156,11 @@ def find_Bragg_disks_CUDA(
                 patt_idx = batch_idx * batch_size + subbatch_idx
                 rx, ry = np.unravel_index(patt_idx, (datacube.R_Nx, datacube.R_Ny))
                 batched_subcube[subbatch_idx, :, :] = cp.array(
-                    datacube.data[rx, ry, :, :]
-                    if filter_function is None
-                    else filter_function(datacube.data[rx, ry, :, :]),
+                    (
+                        datacube.data[rx, ry, :, :]
+                        if filter_function is None
+                        else filter_function(datacube.data[rx, ry, :, :])
+                    ),
                     dtype=cp.float32,
                 )
 
--- a/py4DSTEM/io/importfile.py
+++ b/py4DSTEM/io/importfile.py
@@ -75,7 +75,7 @@ def import_file(
         "gatan_K2_bin",
         "mib",
         "arina",
-        "abTEM"
+        "abTEM",
         # "kitware_counted",
     ], "Error: filetype not recognized"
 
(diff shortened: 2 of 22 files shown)


gvarnavi:
Thanks @sezelt - looks good to me!

Quick-note re: how BO plays w/ the new device/storage memory management: 

The current default behavior for [`clear_fft_cache=True`](https://github.com/py4dstem/py4DSTEM/blob/dev/py4DSTEM/process/phase/singleslice_ptychography.py#L131) is to clear the cupy FFT plans at the end of preprocess/reconstruct. This might make less sense for BO since we're running `reconstruct` over and over again.

The user can always specify `clear_fft_cache=False` in `init_args` to disable this, but perhaps it might make more sense to add this to [`_set_optimizer_defaults`](https://github.com/py4dstem/py4DSTEM/blob/dev/py4DSTEM/process/phase/parameter_optimize.py#L644) and simply call [`ptycho.clear_device_mem(ptycho._device,True)`](https://github.com/py4dstem/py4DSTEM/blob/dev/py4DSTEM/process/phase/phase_base_class.py#L117) during the `finally` clause.

sezelt:
I'm curious, how much time does it actually take to generate the plan, and how much memory does it use?

gvarnavi:
I think it's actually not insignificant, since it's of the order of `(max_batch_size,Qx,Qy)`.   

You can get a feel on how slow it is by seeing how much slower the first ptycho iteration is from subsequent iterations with `clear_fft_cache=True`, and test the memory usage with the notebook I attached [here](https://github.com/py4dstem/py4DSTEM/pull/597#issuecomment-1889864637).

sezelt:
I added `clear_fft_cache=False` to the default optimizer args. Calling `ptycho.clear_device_mem` is trickier because the ptycho object is wrapped by the optimizer function, so it's not in scope at the level of `optimizer.optimize`. I think the better solution (and one which potentially will fix other GPU memory usage issues I've been having in my notebooks recently even without using the optimizer) is for the `ptycho` object to clear the plans on its destruction.

sezelt:
There also seems to be something odd happening with the `black` style check, since it wants to reformat 21 files unrelated to this PR. Perhaps the version of black on the GH action has changed?

sezelt:
So it does seem a new version of `black` was released, and that has led to new rules in some places. I've applied the new formatting in this PR (the only functional changes are the ones originally described).

sezelt:
@gvarnavi I made the change you asked for, but haven't tested everything. Is the current approach to your liking?

gvarnavi (approved):
lgtm, thanks @sezelt!

PR #605, opened 2024-02-20
Adding Friedel center finding to ptycho reconstructions

cophus:
This PR adds a Friedel center finding routine.  I tried adding it to single slice ptycho, but the implementation needs some reworking to make it work with device = 'gpu', and the other flavours of ptycho recon.


cophus:
Here is a testing notebook for the center finding:
https://drive.google.com/file/d/1nxxMSJcgpfNiXAUOv2vSMkcgWDvKGYNx/view?usp=sharing
image

PR #607, opened 2024-02-22
Friedel origin finder, various FEM tools

cophus:
This PR adds the new Friedel origin finder, with and without masking. It also adds new FEM functionality, including local variance (per pattern  in each annular ring), and various small plotting fixes.


cophus:
GPU support now added - testing notebook updated:

https://drive.google.com/file/d/1fbeVyDuQOPLlUWHRdAPnDjKebDKEhJC9/view?usp=sharing

py4DSTEM/process/polar/polar_analysis.py
```diff
@@ -596,18 +600,298 @@ def calculate_FEM_local(
...
+        disable=not progress_bar,
+    ):
+        im = self.data[rx, ry]
+
+        if use_median:
+            im_mean = np.ma.mean(im, axis=0)
+            im_var = np.ma.mean((im - im_mean) ** 2, axis=0)
+        else:
+            im_mean = np.ma.median(im, axis=0)
+            im_var = np.ma.median((im - im_mean) ** 2, axis=0)
```
sezelt, line 646: Are these flipped?
```suggestion
        if use_median:
            im_mean = np.ma.median(im, axis=0)
            im_var = np.ma.median((im - im_mean) ** 2, axis=0)
        else:
            im_mean = np.ma.mean(im, axis=0)
            im_var = np.ma.mean((im - im_mean) ** 2, axis=0)
```

py4DSTEM/process/polar/polar_analysis.py
```diff
@@ -596,18 +600,298 @@ def calculate_FEM_local(
...
+        self._datacube.shape[1],
+        desc="Annular symmetry",
+        unit=" probe positions",
+        disable=not progress_bar,
+    ):
+        # Get polar transformed image
+        im = self.transform(
+            self.data_raw.data[rx, ry],
+            # returnval = 'zeros',
+        )
```
sezelt, line 744: Is there a reason to repeat the transformation here, since this could operate on a PolarDatacube that would already be transformed once ahead of time?

sezelt (changes requested):
Looks good overall, there is just one place where I think you have mean/median mixed up with respect to the flags and I wanted to check that's right. I also pushed some tiny cleanup changes

py4DSTEM/process/polar/polar_analysis.py
cophus: Good catch!   Yes it was flipped - I think I changed from use_mean to use_median since I think the default assumption is mean.

py4DSTEM/process/polar/polar_analysis.py
cophus: PolarDataCube currently generates the images "on demand" rather than precomputing them.

sezelt (approved):

PR #610, opened 2024-02-26
quick fix for probe masking

cophus:
The default threshold for the probe intensity was 0.2 - this fails for counted data where the intensity is <<1.  The default should be 0.0, especially if we are providing a mask in diffraction space.


sezelt (approved):
This also contains changes other than modifying the threshold, did you want to PR those in as well? 
Either way I'm good with the edits, just want to check and make sure

bsavitzky (approved):

PR #611, opened 2024-02-28
Remove duplicate make_fourier_coords_2D

### LINKED ISSUE #564: Duplication of `make_Fourier_coords2D`
**Is your feature request related to a problem? Please describe.**
The `make_Fourier_coords2D` method is duplicated - once in `preprocess` and once in `process`.  

**Describe the solution you'd like**
- Determine if the two methods are identical or not.
- Determine which version should be removed and remove it
- Make sure that anything importing the removed method can still find it.

**Describe alternatives you've considered**
If the method is needed for preprocessing still, it should remain there and be imported into process.  Otherwise, the preprocess version should be removed.

**Additional context**
This was previously used in preprocess - not sure if it still is following refactors

sezelt:
Fixes #564

### DIFF
--- a/py4DSTEM/process/utils/utils.py
+++ b/py4DSTEM/process/utils/utils.py
@@ -11,16 +11,6 @@ from mpl_toolkits.axes_grid1.anchored_artists import AnchoredSizeBar
 import matplotlib.font_manager as fm
 
 from emdfile import tqdmnd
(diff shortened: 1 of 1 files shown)


bsavitzky (approved):

bsavitzky:
Thanks Steve!

PR #616, opened 2024-03-13
Quality of Life Improvements and Minor Bugfixes

gvarnavi:
Minor `phase_contrast` changes from the past month.
I'm updating the tutorial notebooks this week, so it would be great if these can be included in 0.14.10?

The changes are:
- Adding `positions_offset_ang` to ptycho classes initialization (non-square scans utility)
- Adding `force_com_measured` to ptycho classes `preprocess` (speed-up preprocessing)
- Adding `detector_fourier_mask` to ptycho classes `reconstruct` (enabling masked updates)
- Bugfix for `region_of_interest_shape` to actually resample in reciprocal-space
- Adding a `cross_correlate_amplitudes_to_probe_aperture` method to ptycho (enabling better possibly-inline centering of diffraction patterns)


py4DSTEM/preprocess/preprocess.py
```diff
@@ -630,12 +641,28 @@ def resample_data_diffraction(
...
+                grid_mode=True,
+            )
+
+        if conserve_array_sums:
+            output_data = output_data / resampling_factor.prod()
+
+        datacube.data = output_data
         datacube.calibration.set_Q_pixel_size(
-            datacube.calibration.get_Q_pixel_size() / resampling_factor[2]
+            datacube.calibration.get_Q_pixel_size() / resampling_factor[0]
```
cophus, line 665: This will need to be updated once we allow different pixel sizes in qx / qy.

py4DSTEM/preprocess/preprocess.py
```diff
@@ -630,12 +641,28 @@ def resample_data_diffraction(
...
+            datacube.shape[1],
+            desc="Resampling 4D datacube",
+            unit="DP",
+            unit_scale=True,
+        ):
+            output_data[Rx, Ry] = zoom(
+                datacube.data[Rx, Ry].astype(np.float32),
+                resampling_factor,
+                order=1,
+                mode="grid-wrap",
```
cophus, line 656: shouldn't it be `mode = "nearest"`  ?

cophus:
For the Fourier space masking of the updates, do the docstrings state that the mask is True/1 where you want to pass values through, and False / 0 where you want to supress them?  We might want to specify if the convention is the opposite of numpy masked arrays.

cophus (changes requested):
Just a few minor questions / suggestion.

py4DSTEM/preprocess/preprocess.py
gvarnavi: This came from an early discussion with @sezelt where one of his datasets had severe artifacts during bilinear resampling. Although I don't think we had tried "nearest"?

sezelt:
Should the application of the Fourier mask happen before or after the calculation of the error? I think it is more reasonable for the error to represent the difference between the _valid_ intensities and the model rather than the use the unmasked data, especially if there are erroneous pixels that could dominate the error.

gvarnavi:
Indeed, this is one of the reasons I advocated for this [(now commented out) line](https://github.com/py4dstem/py4DSTEM/blob/1afa166259e0367a68f7e522d3c11126a933b9d6/py4DSTEM/process/phase/ptychographic_methods.py#L1564) 🤷🏼 
```python
        fourier_overlap = xp.fft.fft2(overlap)  # * fourier_mask # GV votes to include
```

(although following the resampling changes, if we decide to now include the masking pre-error calculation it should happen after the resampling to make sure the shapes are compatible)

py4DSTEM/preprocess/preprocess.py
cophus: Oh is mode not the edges?  It's the "wrap" part I was worried about here.

gvarnavi: Changed "wrap" to "nearest" for the DP resampling. There's an equivalent function `process.phase.utils.bilinear_resample` which accepts `grid_mode=True` and `mode="grid-wrap"` flags which default to the values above since we use that function to resample corner-centered exit waves.

cophus (approved):
looks good!

PR #617, opened 2024-03-13
Bump gdown to 5.1.0

gvarnavi:
Colaboratory compatibility and other odd bugfixes listed in their release notes.

To reproduce in Colab, this seems to work (while all earlier versions I checked do not):
```python
!pip install "gdown>=5.1.0" py4DSTEM

import py4DSTEM
py4DSTEM.io.gdrive_download(
    id_ = 'https://drive.google.com/uc?id=1C3tXV5BXz0JXbmyu3wTu3NutuIYYsN8A',
    destination = '/content/',
    filename = 'ptycho_MoS2_bin2.h5',
    overwrite=True
)
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(
         "scikit-optimize >= 0.9.0",
         "tqdm >= 4.46.1",
         "dill >= 0.3.3",
-        "gdown >= 4.7.1",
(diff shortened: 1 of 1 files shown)


sezelt (approved):
LGTM

PR #630, opened 2024-03-18
You Can’t Depend On Your Eyes When Your Imagination Is Out of Focus

gvarnavi:
Quick ptycho fix:
- move reset before object_type change

Quick parallax fixes:
- odd padding cropping bugfix
- upsample, fit, correct should return self
- removed Wiener filtering from correct
- correct (?) depth-sectioning


cophus (approved):
Looks correct to me.

PR #631, opened 2024-03-20
Fixing projection 2D thickness

cophus:
My math was all wrong before for the thickness tiling - it should be correct now!

### DIFF
--- a/py4DSTEM/process/diffraction/crystal_ACOM.py
+++ b/py4DSTEM/process/diffraction/crystal_ACOM.py
@@ -2025,6 +2025,9 @@ def calculate_strain(
     tol_intensity: float = 1e-4,
     k_max: Optional[float] = None,
     min_num_peaks=5,
+    intensity_weighting = False,
+    robust = True,
+    robust_thresh = 3.0,
     rotation_range=None,
     mask_from_corr=True,
     corr_range=(0, 2),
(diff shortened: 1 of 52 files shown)


py4DSTEM/process/diffraction/crystal_ACOM.py
```diff
@@ -2098,6 +2123,8 @@ def calculate_strain(
 
     # Loop over all probe positions
     for rx, ry in tqdmnd(
+        # range(220,221),
+        # range(40,41),
```
sezelt, line 2127:
```suggestion
```

py4DSTEM/process/utils/single_atom_scatter.py
```diff
@@ -46,6 +47,34 @@ def electron_scattering_factor(self, Z, gsq, units="A"):
         elif units == "A":
             return fe
 
+    def projected_potential(self, Z, R):
```
sezelt, line 50: What are the units of the returned potential?

sezelt (changes requested):
looks good to me, I just suggested deleting some test lines and clarifying what units the potential are in

py4DSTEM/process/diffraction/crystal_ACOM.py
cophus: Good catch for both.  I tried to figure out the Lobato potential prefactors to properly set the output units, but no luck so far.

sezelt (approved):

PR #639, opened 2024-03-27
Fixing thickness error  in  generate_projejected_potential

shinjandutta:
Earlier version of the code generated potentials where the corners of the image had atomic columns of a different thickness than the ones at the centre of the image. Current version fixes that and now a more uniform image is generated. What remains to be fixed: after a certain threshold for 'thickness_angstroms' the thickness of the atomic columns stops increasing.


shinjandutta:
'Elev' and 'Azim' in CrystalACOM.py were being called before being defined

PR #640, opened 2024-03-27
Thickness of Atomic columns at the edges of the potential images is fixed

shinjandutta:
1. Moved up the code for calculate_correlation_array so as to calculate zone axes without the structure factor calculations in CrystalACOM.py
2. Changed the code for generate_projected_potential in Crystal.py to fix the issue of varying thicknesses of the atomic columns in the generated potential images.


cophus:
Code looks good - we double checked the modifications to orientation plan generation.

cophus (commented):
Looks good!  Projected potentials should be far more accurate now.

cophus (approved):

PR #656, opened 2024-05-09
fix probe norm bug

bsavitzky:
This PR fixes a bug in probe generation, where non-zero pixels outside of the center beam caused normalization issues. 

Note that the solution here makes use of the Probe.measure_disk method instead of the datacube.get_probe_size method which implement similar functionality to avoid breaking changes elsewhere for now. A more comprehensive consolidation of functionality (there is a third method with similar functionality in process.calibration.probe, as well as some redundant metadata storage) will be implemented in v15.  For (upcoming) v14.15+, using Probe.measure_disk is recommended.

### DIFF
--- a/py4DSTEM/braggvectors/probe.py
+++ b/py4DSTEM/braggvectors/probe.py
@@ -4,7 +4,9 @@ import numpy as np
 from typing import Optional
 from warnings import warn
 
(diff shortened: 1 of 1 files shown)


smribet (approved):
Very helpful for noisy data!

PR #662, opened 2024-06-05
Plotting option for flowlines

yaeltsaPost:
Added the option to plot v lines to make it easy to choose radial ranges for orientation histogram.

### DIFF
--- a/py4DSTEM/process/polar/polar_peaks.py
+++ b/py4DSTEM/process/polar/polar_peaks.py
@@ -660,11 +660,28 @@ def plot_radial_peaks(
     qstep=None,
     label_y_axis=False,
     figsize=(8, 4),
+    v_lines=None,
     returnfig=False,
 ):
     """
     Calculate and plot the total peak signal as a function of the radial coordinate.
 
+    q_pixel_units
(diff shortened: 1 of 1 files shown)


py4DSTEM/process/polar/polar_peaks.py
```diff
@@ -741,6 +758,15 @@ def plot_radial_peaks(
...
         ax.tick_params(left=False, labelleft=False)
 
+    if v_lines is not None:
+        y_min, y_max = ax.get_ylim()
+
+        if np.isscalar(v_lines):
+            ax.vlines(v_lines, y_min, y_max, color="g")
+        else:
+            for a0 in range(len(v_lines)):
+                ax.vlines(v_lines[a0], y_min, y_max, color="g")
```
bsavitzky, line 768: I believe the if/else isn't necessary - `ax.vlines(lines,ymin,ymax)` can accept tuples or lists for `lines` - i.e. we can replace these 5 lines with just line 765

bsavitzky (changes requested):
Thanks @yaeltsaPost , the documentation is especially appreciated.  See the one small comment - can you replace that if/else?

PR #663, opened 2024-06-06
hot fix for fit amorphous ring

smribet:
quick fix for passing an initial guess for coefficients for fitting amorphous rings


sezelt (commented):
Could you just reduce the indentation of lines 129-142 instead? The new lines seem to just be a copy, and there's no reason to duplicate if they run regardless of the `if` block

smribet:
Yes!

sezelt (approved):

PR #667, opened 2024-06-29
A little Ruff around the edges

alex-rakowski:
This PR changes the linter for flake8 to Ruff. We should agree on how exactly we want the linter configured before squashing and merging

- adds ruff.toml to configure ruff
- adds ruff.yml to configure and run ruff on only changed files in Github Actions
- updates pre-commit.yml to run black and ruff and removes flake8 
- removes various flake8 config and Github Action files.


CSSFrancis:
Just a comment, The Pre-commit CI is often a little easier to use then creating a github workflow 

https://pre-commit.ci

I like it as you can just format other peoples code by running  "pre-commit.ci autofix" which is helpful if you are trying to enforce things like code quality checks etc.

PR #674, opened 2024-08-11
Phase contrast quality of life improvements

gvarnavi:
Various quality of life improvements for the `phase` module. Namely:
- Pixel dimensions invariant BO metrics
- Optional in-place preprocessing of datacube
- `max_batch_size` option in parallax `preprocess` and `reconstruct` (only useful for shifting cross-correlated VBFs)
- `aligned_bf_image_guess` option in parallax `preprocess` to initialize cross-correlation with
- new `guess_common_aberrations` function which accepts microscope geometry parameters (rotation/transpose), as-well as common aberrations (defocus, stig, coma, spherical), and computes the analytical BF shifts -- using them to compute the aligned (and optionally upsampled) BF image
  - note this returns a static figure, but is easy to make into a widget inside a jupyter notebook
  ```python
  import ipywidgets
  
  def widget_wrapper(rotation_angle_deg, transpose, defocus, upsampling_factor):
      """ wrapper to only pass interactive sliders of interest """
      return parallax.guess_common_aberrations(
          rotation_angle_deg = rotation_angle_deg,
          transpose = transpose,
          defocus = defocus,
          kde_upsample_factor = upsampling_factor,
      )
      
  ipywidgets.interact(
      widget_wrapper,
      rotation_angle_deg = -15,
      transpose = False,
      defocus = 1.5e4,
      upsampling_factor = 1,
  );
  ```

### DIFF
--- a/py4DSTEM/process/phase/magnetic_ptychographic_tomography.py
+++ b/py4DSTEM/process/phase/magnetic_ptychographic_tomography.py
@@ -225,6 +225,7 @@ class MagneticPtychographicTomography(
         padded_diffraction_intensities_shape: Tuple[int, int] = None,
         region_of_interest_shape: Tuple[int, int] = None,
         dp_mask: np.ndarray = None,
+        in_place_datacube_modification: bool = False,
         fit_function: str = "plane",
         plot_probe_overlaps: bool = True,
         rotation_real_space_degrees: float = None,
@@ -266,6 +267,9 @@ class MagneticPtychographicTomography(
             at the diffraction plane to allow comparison with experimental data
         dp_mask: ndarray, optional
             Mask for datacube intensities (Qx,Qy)
+        in_place_datacube_modification: bool, optional
+            If True, the datacube will be preprocessed in-place. Note this is not possible
+            when either crop_patterns or positions_mask are used.
         fit_function: str, optional
             2D fitting function for CoM fitting. One of 'plane','parabola','bezier_two'
         plot_probe_overlaps: bool, optional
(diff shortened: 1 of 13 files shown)


sezelt (changes requested):
For the in-place datacube modification, I think we may want to somehow prevent someone from accidentally preprocessing the same dataset more than once. Perhaps there is a clean way to set some flag on the datacube or add an entry to its metadata to indicate that it has already been modified, and then warn or error if you try to run it again.

I like the blind parallax a lot. It would be nice if the plotting was behind a flag, so that I could use this as the basis of a parallax feature in the GUI. Also, can you really handle transpose just by changing the direction of the rotation? Maybe it's happening at a different stage than I am used to but that just seemed fishy.

gvarnavi:
> For the in-place datacube modification, I think we may want to somehow prevent someone from accidentally preprocessing the same dataset more than once. Perhaps there is a clean way to set some flag on the datacube or add an entry to its metadata to indicate that it has already been modified, and then warn or error if you try to run it again.

To be honest I'm not a huge fan of this feature, but a user working on very large datasets asked for it and figured I'd oblige. I think if the user is explicitly setting `in_place_datacube_modification = True` it should be on the user to make sure they only do it once? e.g. load the data in the same cell or something.

> I like the blind parallax a lot. It would be nice if the plotting was behind a flag, so that I could use this as the basis of a parallax feature in the GUI. 

Sure, I'll add a plot for the plotting. It would be amazing to add this functionality to the GUI!

> Also, can you really handle transpose just by changing the direction of the rotation? Maybe it's happening at a different stage than I am used to but that just seemed fishy.

It suffices to perform these two operations, yes:
1. passively rotate the Fourier-coordinate system by negative theta
```python
        if transpose:
            rotation_angle_deg *= -1
```
and flip the shifts after they're computed:
```python
        if transpose:
            shifts_ang = xp.flip(shifts_ang, axis=1)
```

sezelt:
I see, I just missed the second step of flipping.

sezelt (approved):

PR #676, opened 2024-08-29
Update polar_analysis.py

maclariz:
Updated to allow calculation of rdfs from single points or small areas

Main point was to replace calls to self.radial_mean with defining Ik either as self.radial_mean or areas selected/averaged from self.radial_all

Some minor updates were also needed in plot_background_fits

Some doc strings added

The one weakness in my code is that the rdf from the sub area is then attached to self as representative of whole dataset.  This doesn't worry me for now.  If I want a different sub-area, I can recalculate.  But you may wish to do point by point calculation over entire dataset at some point.

### DIFF
--- a/py4DSTEM/process/polar/polar_analysis.py
+++ b/py4DSTEM/process/polar/polar_analysis.py
@@ -526,8 +526,8 @@ def plot_background_fits(
     ax.set_ylabel("I(k) and Background Fit Estimates")
     ax.set_ylim(
         (
-            np.min(self.radial_mean[self.radial_mean > 0]) * 0.8,
-            np.max(self.radial_mean * self.Sk_mask) * 1.25,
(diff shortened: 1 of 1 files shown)


smribet:
This is great, thanks @maclariz!

Should we update the initial guess for the background coefficients to use Ik as well (line 346/346) of `polar_analysis.py`? 
```
 const_bg = np.min(self.radial_mean) / int_mean
 int0 = np.median(self.radial_mean) / int_mean - const_bg
```

maclariz:
You're absolutely right.  These were initial guess for an optimisation so it probably didn't make any difference when I ran it since it probably found the minimum anyway.  However, this is probably closer for sub-areas and better practice.

smribet:
Awesome, thanks!

smribet (approved):

PR #677, opened 2024-08-29
numpy < 2

smribet:
Changing set-up file (for v 0.14 only) to avoid numpy 2 install errors

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,11 +23,11 @@ setup(
     keywords="STEM,4DSTEM",
     python_requires=">=3.10",
     install_requires=[
-        "numpy >= 1.19",
+        "numpy >= 1.19, < 2.0",
         "scipy >= 1.5.2",
         "h5py >= 3.2.0",
         "hdf5plugin >= 4.1.3",
-        "ncempy >= 1.8.1",
+        "ncempy >= 1.8.1, < 1.11.2",
         "matplotlib >= 3.2.2",
         "scikit-image >= 0.17.2",
         "scikit-learn >= 0.23.2, < 1.5",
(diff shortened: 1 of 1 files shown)


mlhenderson:
@smribet Yes to numpy < 2.0.  Is the ncempy restriction also due to numpy?

mlhenderson (approved):

smribet:
Yes I think I made a mistake specifying the version though. Should it be <= 1.11.2?

mlhenderson:
<= 1.11.2 would be better, yes.  There are other changes for 1.11.2 besides the numpy restriction that should be picked up, which you would not get from 1.11.1.

mlhenderson (changes requested):
we should also restrict pymatgen to < 2024.8.9, since that is pinned to numpy 2.0.0

mlhenderson:
I did some local py4DSTEM testing using numpy 2, and emdfile has at least one incompatibility with numpy 2 currently.  It makes sense to put some version limits for py4DSTEM numpy install, at least temporarily, until the problems related to dependencies can be addressed.

bsavitzky:
Thanks Matt - what incompatibility with emdfile did you find?

mlhenderson:
I filed an issue with emdfile, so far one error related to usage of numpy.string_

https://github.com/py4dstem/emdfile/issues/2

bsavitzky:
Great, thanks - fixed and pushed to a new branch `np2` for numpy 2
compatibility fixes.  Will leave the branch open for now while you're
testing

smribet:
@bsavitzky it sounds like we should pin emdfile as well?

bsavitzky:
Once we've ensured numpy 2 compatibility I'll merge and version emdfile.
Then we could simply bump the version dependency in py4DSTEM but leave the
>=. I think this is better than pinning in this situation but am curious -
do you think pinning would be preferable?

smribet:
Sounds good!

bsavitzky (approved):

PR #681, opened 2024-09-04
DDF patch 4

maclariz:
So, as discussed, I added some extra functionality to pointslist_to_array so it now adds qr and qphi by default, although this can be turned off

I then also added the first radial imaging function, DDF_radial_image

Hope you like them


bsavitzky (approved):
Looks good.

maclariz:
Thanks Ben,
A few further functions using polar definitions will follow in due course, but not quite in final form.  I'm now taking a few days off so will get back to it in due course.

bsavitzky:
Sounds good, thanks Ian.  We've just released the code with your recent PRs
included (v0.14.17, up on pip now, should be up on conda by the end of the
day or tomorrow).  We can get your additional functions included in the
next version :)

PR #688, opened 2024-09-19
Update digital_dark_field.py

maclariz:
Radius filtering separated as function, and radius-azimuth imaging added as separate function

### DIFF
--- a/py4DSTEM/process/diffraction/digital_dark_field.py
+++ b/py4DSTEM/process/diffraction/digital_dark_field.py
@@ -552,8 +552,8 @@ def DDF_radial_image(points_array_w_rphi, radius, Rshape, tol=1):
 
     if Rshape is None:
         Rshape = (
-            np.max(np.max(points_array[:, 3])).astype("int") + 1,
-            np.max(np.max(points_array[:, 4])).astype("int") + 1,
+            np.max(np.max(points_array_w_rphi[:, 3])).astype("int") + 1,
+            np.max(np.max(points_array_w_rphi[:, 4])).astype("int") + 1,
         )
 
     radial_filtered_points_array = radial_filtered_array(
@@ -603,8 +603,8 @@ def DDFradialazimuthimage(points_array_w_rphi, radius, phi0, phi1, Rshape, tol=1
     """
     if Rshape is None:
         Rshape = (
-            np.max(np.max(points_array[:, 3])).astype("int") + 1,
-            np.max(np.max(points_array[:, 4])).astype("int") + 1,
+            np.max(np.max(points_array_w_rphi[:, 3])).astype("int") + 1,
+            np.max(np.max(points_array_w_rphi[:, 4])).astype("int") + 1,
         )
 
     radial_filtered_points_array = radial_filtered_array(
@@ -623,8 +623,8 @@ def DDFradialazimuthimage(points_array_w_rphi, radius, phi0, phi1, Rshape, tol=1
     )
     radiusazimuthimage = np.zeros(shape=Rshape)
 
-    for i in range(h):
-        for j in range(w):
+    for i in range(Rshape[0]):
+        for j in range(Rshape[1]):
             radiusazimuthimage[i, j] = np.where(
                 np.logical_and(
                     rphi_filtered_points_array[:, 3] == i,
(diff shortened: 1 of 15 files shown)


py4DSTEM/process/diffraction/digital_dark_field.py
```diff
@@ -494,47 +494,143 @@ def DDFimage(points_array, aperture_positions, Rshape=None, tol=1):
     return image
 
 
-def DDF_radial_image(points_array, radius, Rshape, tol=1):
+def radial_filtered_array(points_array_w_rphi, radius, tol=1):
```
bsavitzky, line 497: This method now returns a filtered array, and not a digital dark field image - can you update the first line of the docstring to reflect that?

py4DSTEM/process/diffraction/digital_dark_field.py
```diff
@@ -494,47 +494,143 @@ def DDFimage(points_array, aperture_positions, Rshape=None, tol=1):
...
+                radial_filtered_points_array[:, 2],
                 0,
             ).sum()
 
     return radialimage
+
+
+def DDFradialazimuthimage(points_array_w_rphi, radius, phi0, phi1, Rshape, tol=1):
+    """
+    Calculates a Digital Dark Field image from a list of detected diffraction peak positions in a points_array matching a specific qr radius, within a defined matching tolerance
```
bsavitzky: This docstring should be updated to explain that the output image is from the selected azimuthal range only.

py4DSTEM/process/diffraction/digital_dark_field.py
```diff
@@ -494,47 +494,143 @@ def DDFimage(points_array, aperture_positions, Rshape=None, tol=1):
...
+    )
+
+    rphi_filtered_points_array = np.delete(
+        radial_filtered_points_array,
+        np.where(
+            np.logical_or(
+                radial_filtered_points_array[:, 6] < phi0,
+                radial_filtered_points_array[:, 6] > phi1,
+            )
+        ),
```
bsavitzky: This logic doesn't look like it will be robust to [phi0,phi1] spanning the branchcut.  Is that right or have I missed something?

bsavitzky (changes requested):
Hi @maclariz - thanks for the PR and sorry for the delay.  Please see the comments - I've requested two small docstring edits and one code edit for azimuthal ranges spanning a branchcut.  Happy to get this pulled in once that's all squared - cheers :)

py4DSTEM/process/diffraction/digital_dark_field.py
maclariz: All changes made.  Docstrings edited.  Line 619 ">" changed to ">="

maclariz:
Now updated.  Initially failed on code style but has been run through black and should now pass.

py4DSTEM/process/diffraction/digital_dark_field.py
bsavitzky: Thanks @maclariz .  I think this still is not robust to branchcuts, however - by which I mean, if you've conditioned your data such that `phi` is in [0,360] (for example), then the 20 degree range [-10,10] (or equivalently [350,10]) can't be specified with this code.  You can sum virtual images from two ranges as a workaround, though handling it here would be nicer.  Up to you - its a quality of life change, there's certainly nothing incorrect here, just a small suggestion.  I'll lint and then pull this in as-is and if you'd like to change that later you always can :)

bsavitzky:
Hi @maclariz - i've just pushed commits updating the black formatting so that the PR will pass the linter.  This adds changes to a number of other files that were not passing linting, which looks like it has to do with a version update to black.  Sorry about that and please don't worry about the other files elsewhere in the package that are updated - this is all black formatting changes.

Aside from my branchcut comment, all looks good to me now.  I see you've marked the patch as a draft, so I'll wait on merging until I hear otherwise from you.

py4DSTEM/process/diffraction/digital_dark_field.py
```diff
@@ -494,47 +494,145 @@ def DDFimage(points_array, aperture_positions, Rshape=None, tol=1):
...
+        points_array_w_rphi, radius, tol
+    )
+
+    rphi_filtered_points_array = np.delete(
+        radial_filtered_points_array,
+        np.where(
+            np.logical_or(
+                radial_filtered_points_array[:, 6] < phi0,
+                radial_filtered_points_array[:, 6] >= phi1,
+            )
```
bsavitzky, line 622: this still does not handle the branchcut issue (i.e. for data conditioned such that `phi` is e.g. in the range [0,360], you can't specify a range like [350,10] i.e. [-10,10] with this code).  Let's call this is a quality-of-life / interface issue and you can decide if you want to update this now or later at your discretion; for now I'm ok pulling the code in if you're ok with this small interface issue

bsavitzky (approved):

PR #673, opened 2024-07-12
Adjust scalebar logic to combine any specified parameters with defaults

edwinsupple:
Fix for issue #672

Issue #672, opened 2024-07-12
Show scalebar parameters don't take effect if ar.calibration is not None


edwinsupple:
**Describe the bug**
when using `py4DSTEM.visualize.show`, a dictionary passed to `scalebar` is ignored if the plotted image has a calibration parameter (e.g. virtual bright field images)

**To Reproduce**
```
import py4DSTEM
from py4DSTEM.visualize import show
import numpy as np

datacube = py4DSTEM.DataCube(np.random.rand(10,10,128,128))

datacube.calibration.set_Q_pixel_units('mrad') # arbitrary values to fill out calibration
datacube.calibration.set_Q_pixel_size(0.03)
datacube.calibration.set_R_pixel_units('nm')
datacube.calibration.set_R_pixel_size(3)

bf1 = datacube.get_virtual_image(mode='circle', geometry = ((64,64), 30))

show(bf1) # displays the default scalebar
show(bf1, scalebar={'position':'bl', 'color':'k'})  # scalebar still in same position and color as default

```
**Expected behavior**
Would expect the default scalebar parameters to take effect, except for any that are passed to the function in the scalebar dictionary.

**py4DSTEM version** 
0.14.15

**Python version**
3.10.4
**Operating system** 
Windows 

**Screenshots**
If applicable, could you add screenshots to help explain your problem?
current functionality, showing parameter passed to `scalebar` as axis title:
![scalebar_functionality-old](https://github.com/user-attachments/assets/b64ea1f7-c2bd-403b-9991-b736ab48bcd4)

expected behavior:
![scalebar_functionality-new](https://github.com/user-attachments/assets/11b278b2-385d-4ffd-a310-dfb02d14efba)

bsavitzky:
Thank you Edwin!

PR #690, opened 2024-10-01
Fix for show_complex interaction with scalebar dict

edwinsupple:
See #673 and #689. `show_complex` incorrectly assigned None to `figsize` when popping it from the kwargs resulting in broken plotting due to interaction with `scalebar`, it now assigns a default figsize as is typical for other show type functions. I did not find the same issue in the other show functions. @smribet please let me know if you have a case where this is still broken.

### DIFF
--- a/py4DSTEM/visualize/show.py
+++ b/py4DSTEM/visualize/show.py
@@ -186,7 +186,7 @@ def show(
         and
 
             >>> show(dp, calibration=calibration, scalebar={'length':0.5,'width':2,
-                                                       'position':'ul','label':True'})
+                                                       'position':'ul','label':True})
 
         will display a more customized scalebar.
 
@@ -316,7 +316,7 @@ def show(
         if returnfig==False (default), the figure is plotted and nothing is returned.
         if returnfig==True, return the figure and the axis.
     """
-    if scalebar is True:
+    if scalebar is True or scalebar is None:
         scalebar = {}
 
     # Alias dep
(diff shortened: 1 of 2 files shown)


gvarnavi:
Thanks @edwinsupple for the PR. 

If I recall correctly #673 broke a number of other downstream visualization functions used in the `phase` module (adding multiple scalebars etc). I can try and reproduce those bugs later this week.

Taking a step back, it's not obvious to me popping kwargs is the way to go? A simpler approach for #672 could perhaps be to check whether the user passed any scalebar options, and if so and the array to be plotted has their own calibrations, to simply extract the array (`bf.data`) and plot that?

edwinsupple:
@gvarnavi there are a lot of instances where `figsize` gets set via something like `figsize = kwargs.pop("figsize", (5, 5))`. For example `_visualize_last_iteration` and `_visualize_all_iterations` in dpc.py. 

Looking at this more, I think the issue may be that my initial PR did not handle None correctly. I was focused on the first line in the docstring about adding a scalebar when a DiffractionSlice or RealSlice is passed and missed the bit about not adding a scalebar when an array is passed. `scalebar=None` was changed to `scalebar={}` at the start of `show` and then lines 757-772 do happen when they shouldn't.

I will go back and work on this more, if you have a test case or two I can try let me know.

gvarnavi:
Indeed, I added those haha -- the purpose of that is to allow compatibility with different function call structures and impose opinionated defaults in the downstream functions.

To be quite honest, the objection here is on the way kwargs get handled in `show`, which of-course predates #673. We have had many many discussions internally that `show` has become too unwieldy and needs to be refactored.. 

Which is why the suggestion was a minimally-invasive "hack" to achieve your effect w/o touching the entangled kwargs mess.

PR #696, opened 2024-11-20
Update digital_dark_field.py

maclariz:
Fixed one error that causes incorrect calculation of DDFimage if the pointsarray has 7 columns, by simply changing 2: to 2:5 in the calculation


maclariz:
@cophus @bsavitzky Just a simple fix of one digit to clear up a calculation error if the 7 column array is used

bsavitzky (approved):

PR #700, opened 2024-12-08
Update WK_scattering_factors.py

ZacharyVarley:
Relevant WK computation in others.f90 in Prof. Marc De Graef's EMsoft reads: 

```
DO J=1,4
  FPHON = FPHON + A1(J)*A1(J)*(DEWA * RI1(B1(J),B1(J),G) - RI2(B1(J),B1(J),G,UL))
  DO I=1,J-1
   FPHON = FPHON + 2.*A1(J)*A1(I)*(DEWA * RI1(B1(I),B1(J),G)-RI2(B1(I),B1(J),G,UL))
```

When J=1:
The inner loop condition is DO I=1,0 (since J-1 = 0). In Fortran, when the upper bound is less than the lower bound, the loop is not executed at all. So for J=1, only the first FPHON calculation happens, the inner loop is skipped. In the current py4DSTEM code, the inner loop is executed for ii = jj = 0 making the diagonal computed a second time with a factor of 2.0, for an overall prefactor of 3.0 instead of 1.0 when ii equals jj. If I am missing something and this python code has already been verified against others.f90, then completely ignore this.

### DIFF
--- a/py4DSTEM/process/diffraction/WK_scattering_factors.py
+++ b/py4DSTEM/process/diffraction/WK_scattering_factors.py
@@ -174,14 +174,9 @@ def compute_WK_factor(
         B1 = B / (4.0 * np.pi) ** 2
 
         for jj in range(4):
-            Fphon += (
-                A1[jj]
-                * A1[jj]
-                * (DWF * RI1(B1[jj], B1[jj], G) - RI2(B1[jj], B1[jj], G, UL))
-            )
             for ii in range(jj + 1):
                 Fphon += (
-                    2.0
+                    (2.0 if jj != ii else 1.0)
                     * A1[jj]
                     * A1[ii]
                     * (DWF * RI1(B1[ii], B1[jj], G) - RI2(B1[ii], B1[jj], G, UL))


sezelt:
Good catch on the discrepancy. It seems we do differ from the relevant snippet in EMSoftLib:
https://github.com/EMsoft-org/EMsoft/blob/2c071d6975265a423b1e83cf5436a8995293c1d3/Source/EMsoftLib/others.f90#L210-L236

I haven't yet had a chance to test this update, but I think we can also fix this more cleanly by just fixing the range limits:
```python
for jj in range(4):
    for ii in range(jj):
        [...]
```

Curiously, during initial implementation I used to have yet another different bounds (both of which probably resulted from misunderstanding the Fortran behavior): https://github.com/py4dstem/py4DSTEM/commit/a74d2fb4de03e75fa290f3b6f30cc3bc95601bc5#diff-652c64a84248c41f4cc70b79f26dd9b2f72223f4079c9a52668e886ebc2bf820R190

ZacharyVarley:
I have written a completely vectorized version (over accelerating voltage, Z & sigma pairs per atom type, and g) in PyTorch and I will try to port it to numpy / scipy for you all at some point. It might be worth putting some sort of disclaimer for those who have used these WK computations thus far as the delta is significant.

sezelt (approved):

PR #701, opened 2024-12-29
New year, new arina data

smribet:
A few small updates to the Arina reader: 
- default square scan
- optional transpose to accommodate new and old files
- metadata reading for new files

 Example [here](https://drive.google.com/file/d/10EfoZWNQtFL5rXvh3Pc3Ad7wEXndQxOI/view?usp=sharing).

### DIFF
--- a/py4DSTEM/io/filereaders/read_arina.py
+++ b/py4DSTEM/io/filereaders/read_arina.py
@@ -7,13 +7,15 @@ from py4DSTEM.preprocess.utils import bin2D
 
 def read_arina(
     filename,
-    scan_width=1,
+    scan_width=None,
     mem="RAM",
     binfactor: int = 1,
     dtype_bin: float = None,
+    fix_transpose = False,
     flatfield: np.ndarray = None,
     median_filter_masked_pixels_array: np.ndarray = None,
     median_filter_masked_pixels_kernel: int = 4,
+
 ):
     """
     File reader for arina 4D-STEM datasets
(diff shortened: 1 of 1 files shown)


py4DSTEM/io/filereaders/read_arina.py
```diff
@@ -90,20 +100,39 @@ def read_arina(
...
         )
-    )
 
     if median_filter_masked_pixels_array is not None and binfactor == 1:
         datacube = datacube.median_filter_masked_pixels(
             median_filter_masked_pixels_array, median_filter_masked_pixels_kernel
         )
 
+    try:
+        with h5py.File(f"{filename[:-10]}.h5", "r") as f:
```
gvarnavi, line 128: let's add a comment here to make sure we don't forget that `[:-10]` comes from dropping the "_master" suffix?

py4DSTEM/io/filereaders/read_arina.py
```diff
@@ -7,10 +7,11 @@
 
 def read_arina(
     filename,
-    scan_width=1,
+    scan_width=None,
     mem="RAM",
     binfactor: int = 1,
     dtype_bin: float = None,
+    fix_transpose=False,
```
gvarnavi, line 14: It would be great if this default could `None`, and we could (reliably) detect whether the file is using the old/new ARINA format and set a default value accordingly if not passed.

gvarnavi (approved):
lgtm, added two minor comments but will go ahead and approve regardless.

PR #704, opened 2025-01-23
Update WK_scattering_factors.py

ZacharyVarley:
Presumed typo changed the cut-off for WEKO argument from the correct value 0.1 to the incorrect value 1.0. This returns the value to 0.1 to match WK's F77 implementation ported to F90 by MDG:

https://github.com/EMsoft-org/EMsoftOO/blob/develop/Source/EMsoftOOLib/mod_others.f90#L191

"IF (ARGU .LT. .1) THEN"

0.1 shows no artifacts in the real part (plot from my private PyTorch repo):
wk_scatter_with_0 1_no_artifacts

and 1.0 shows clear artifacts in the real part (plot from my private PyTorch repo)
wk_scatter_with_1 0_artifacts

### DIFF
--- a/py4DSTEM/process/diffraction/WK_scattering_factors.py
+++ b/py4DSTEM/process/diffraction/WK_scattering_factors.py
@@ -83,7 +83,7 @@ def compute_WK_factor(
     WK = np.zeros_like(S)
     for i in range(4):
         argu = B[i] * S**2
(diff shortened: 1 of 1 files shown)


sezelt (approved):
Another very good catch–thank you!

PR #706, opened 2025-01-29
Small typo, docstring update, and fixed a bug.

tcpekin:
`size_marker` wasn't doing anything in the `plot_structure` function when I was testing it. It looks like `radius` was hardcoded with a magic number. I've just slotted in `size_marker`, added a new magic number, and updated the docstring.

It seems to work for me, but I haven't extensively tested it with a variety of structures, just one. 

Also, why is this code written like this, instead of using the commented out `ax.scatter` method? I was certainly confused.

### DIFF
--- a/py4DSTEM/process/diffraction/crystal_viz.py
+++ b/py4DSTEM/process/diffraction/crystal_viz.py
@@ -150,7 +150,7 @@ def plot_structure(
 
     # atoms
     ID_all = np.unique(ID)
-    if occ is None:
+    if np.all(np.asarray(occ) == 1):
         for ID_plot in ID_all:
             sub = ID == ID_plot
             ax.scatter(
--- a/py4DSTEM/process/diffraction/crystal_viz.py
+++ b/py4DSTEM/process/diffraction/crystal_viz.py
@@ -30,7 +30,7 @@ def plot_structure(
     camera_dist: Optional[float] = None,
     show_axes: bool = False,
     perspective_axes: bool = True,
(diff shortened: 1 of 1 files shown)


sezelt:
This part of the code is for structures with partial occupancy, and it displays the atoms as little pies broken up by the occupancy fraction on each site. For structures with all sites fully occupied, it uses `ax.scatter`.

sezelt (approved):

tcpekin:
@sezelt I'm not sure this code ever hits the `scatter` call to be honest. I am testing with https://legacy.materialsproject.org/materials/mp-81/ (pure gold) and get the following occupancy when I drop a print statement right before the plotting of the atoms. 

image
 

```python
structure = get_mp_structure('mp-81')
crystal = py4DSTEM.process.diffraction.Crystal.from_pymatgen_structure(
    structure=structure,
)

fig, ax = crystal.plot_structure(
    returnfig=True,
    size_marker=200,
    zone_axis_lattice=[1,1,0],
    perspective_axes=True,
)
```


So I think the check of `if occ is None:` will never execute. That's where my confusion stems from. If I remember, I'll submit a small bugfix.

sezelt:
I think you're right and there's an incorrect assumption, as the `Crystal` initializer will always set `self.occupancy` and thus the `if occ is None:` branches are never taken. Instead we should test for any element of `occupancy` being non-one.

tcpekin:
The latest commit should fix it, but I didn't test it to be honest. It may have an extra cast to an array, but that's probably fine.

Issue #710, opened 2025-03-14
Add Numpy 2 compatibility


cmeyer:
We would like to use py4DSTEM in Python environments with numpy 2.x.

The core packages seem to all be available to be used with numpy 2.x. Tested with this command:

`conda create -n trial_dependencies -c conda-forge numpy scipy hdf5 hdf5plugin ncempy matplotlib scikit-image scikit-learn scikit-optimize tqdm dill gdown dask distributed emdfile mpire threadpoolctl pylops colorspacious`

Some of the extensions dependencies may not yet available on conda-forge, but are available on PyPI (tensorflow requires 2.18 for numpy 2 compatibility, 2.18 isn't available using conda-forge on macOS arm64 yet). I'm on a macOS machine right now so I couldn't try everything (i.e. cupy).

sezelt:
When we restricted it <2.0 in #677, the reason was an incompatibility in emdfile, but that seems to have been addressed: https://github.com/py4dstem/emdfile/commit/3cdf83ec1d58fcb9564fb90377e0446f61576813 now. Some incompatibilities remain in the main repo, but hopefully these are only the trivial ones from deprecated aliases.

ericpre:
The conda-forge feedstock still have the numpy<2 pin, can it be removed?

PR #711, opened 2025-03-14
Fix numpy 2 deprecated aliases

sezelt:
Remove the use of aliases deprecated by numpy 2 by running `ruff check py4DSTEM --select NPY201 --fix` and update `setup.py` to no longer restrict `numpy < 2.0`.

Based on the [migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html) I don't anticipate any further problems within py4DSTEM from using numpy 2. 

Addresses #710.


sezelt:
Closing because of mismatched history

PR #712, opened 2025-03-14
Fix numpy 2 deprecated aliases

sezelt:
Remove the use of aliases deprecated by numpy 2 by running `ruff check py4DSTEM --select NPY201 --fix` and update `setup.py` to no longer restrict `numpy < 2.0`.

Based on the [migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html) I don't anticipate any further problems within py4DSTEM from using numpy 2. 

Addresses #710.

### DIFF
--- a/py4DSTEM/io/legacy/legacy13/v13_emd_classes/io.py
+++ b/py4DSTEM/io/legacy/legacy13/v13_emd_classes/io.py
@@ -127,42 +127,42 @@ def Metadata_to_h5(metadata, group):
         # None
         if v is None:
             v = "_None"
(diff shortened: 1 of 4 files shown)


smribet:
I think we should also unpin ncempy

sezelt:
Yes, that can be unpinned. With this change, pip will install numpy 2 so everything in the requirements is compatible

smribet (approved):
Ok! It looks good to me, thanks! I'll approve, but wait until next week to merge in case anyone else wants to weigh in.

PR #726, opened 2025-07-08
Added ability to do segmented peak detection on polar transform of DP

NJMarchese:
- Added new function to allow for different peak detection parameters for different q-spacings on the diffraction pattern. 
- Also enabled the ability to pass a realspace mask onto find_peaks. 
- Added an error catch onto the curve_fit of flowlines to give a more informative error message.
- Added kwargs.get method to find_peaks_single_pattern so that the figure and axes are passed correctly. Before wouldn't handle correctly, and if figax was passed it would give an error saying there were multiple values for figax (the default in the code and the one passed in kwargs)
- Added vmin and vmax kwargs pass to show_amorphous_ring_fit
- Added ability to return data for plot_radial_peaks

### DIFF
--- a/py4DSTEM/process/polar/polar_datacube.py
+++ b/py4DSTEM/process/polar/polar_datacube.py
@@ -115,6 +115,9 @@ class PolarDatacube:
         plot_radial_background,
         model_radial_background,
         make_orientation_histogram,
+        # Added
+        find_peaks_single_pattern_segmented,
+        find_peaks_segmented,
     )
 
     # sampling methods + properties
--- a/py4DSTEM/process/polar/polar_datacube.py
+++ b/py4DSTEM/process/polar/polar_datacube.py
@@ -115,7 +115,6 @@ class PolarDatacube:
         plot_radial_background,
         model_radial_background,
         make_orientation_histogram,
-        # Added
         find_peaks_single_pattern_segmented,
         find_peaks_segmented,
     )
--- a/py4DSTEM/process/polar/polar_peaks.py
+++ b/py4DSTEM/process/polar/polar_peaks.py
@@ -311,8 +311,8 @@ def find_peaks_single_pattern(
         t = np.linspace(0, 2 * np.pi, 180 + 1)
         ct = np.cos(t)
         st = np.sin(t)
-        if kwargs.get('figax'):
-            fig, ax = kwargs.pop('figax')
+        if kwargs.get("figax"):
+            fig, ax = kwargs.pop("figax")
         else:
             fig, ax = plt.subplots(figsize=figsize)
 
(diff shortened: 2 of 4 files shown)


py4DSTEM/visualize/vis_special.py
```diff
@@ -159,6 +159,10 @@ def show_amorphous_ring_fit(
         return_intensity_range=True,
         **kwargs,
     )
+    if kwargs.get("vmax") is not None:
+        vmax = kwargs.pop("vmax")
```
smribet: I think if you want to do this you need add vmin and max as flags to the function

py4DSTEM/process/polar/polar_datacube.py
```diff
@@ -115,6 +115,9 @@ def __init__(
         plot_radial_background,
         model_radial_background,
         make_orientation_histogram,
+        # Added
```
smribet: Please remove any commented out code that's specific to your PR

py4DSTEM/process/polar/polar_peaks.py
```diff
@@ -321,7 +325,8 @@ def find_peaks_single_pattern(
         ax.scatter(
             peaks_polar["qr"],
             peaks_polar["qt"],
-            s=peaks_polar["intensity"] * plot_scale_size,
+            # s=peaks_polar["intensity"] * plot_scale_size,
+            s=10,
             marker="o",
```
smribet, line 329: In general, I'd suggest leaving this more flexible in case someone has different plot sizes than you

py4DSTEM/process/polar/polar_peaks.py
```diff
@@ -472,6 +760,153 @@ def find_peaks(
...
+    threshold_prom_annular=None,
+    threshold_prom_radial=None,
+    remove_masked_peaks=False,
+    scale_sigma_annular=0.5,
+    scale_sigma_radial=0.25,
+    progress_bar=True,
+    kwargs_pass=None,
+):
+    """
+    Peak detection function for polar transformations. Loop through all probe positions,
```
smribet, line 784: missing doc strings...especially for your new contributions with kwargs_pass, it'd be helpful to add detailed docs here!

py4DSTEM/process/polar/polar_peaks.py
```diff
@@ -472,6 +760,153 @@ def find_peaks(
         self.background_radial_mask[rx, ry] = sig_bg_mask
 
 
+def find_peaks_segmented(
+    self,
+    mask=None,
+    mask_real=None,
```
smribet, line 765: Can this be added to the other find_peaks function too?

smribet:
Hi @NJMarchese, thanks for contributing these features! I have a few small suggestions to your code.

py4DSTEM/process/polar/polar_peaks.py
NJMarchese: Returned to the original line of code for marker size

py4DSTEM/visualize/vis_special.py
NJMarchese: Changed to add vmin and vmax as flags for the function input, but also had to change other lines of code since variable with names vmin and vmax assigned in line 154.

py4DSTEM/process/polar/polar_peaks.py
NJMarchese: Maybe. When I looked at it the first time it seemed like it would break a bunch of things, so made it separate initially to keep everything in tact. I can take another look.

py4DSTEM/process/polar/polar_peaks.py
NJMarchese: I put in some doc strings, but as this is built off the original find_peaks function and the docstrings for that function is what is in find_peaks_segmented, I'm not sure if it's best for me to fill in those belonging to the original function. I added docstrings for the parameters I added for find_peaks_segmented.

PR #732, opened 2025-09-23
Updating cluster module and ACOM for cluster based strain mapping

serinlee1065:
updating cluster module and ACOM to use cluster dataset for strain mapping

### DIFF
--- a/py4DSTEM/process/utils/cluster.py
+++ b/py4DSTEM/process/utils/cluster.py
@@ -15,22 +15,49 @@ class Cluster:
     def __init__(
         self,
         datacube,
+        r_space_mask,
     ):
         """
         Args:
-            datacube (py4DSTEM.DataCube):            4D-STEM data
-
+            datacube (py4DSTEM.DataCube):         4D-STEM data
+            r_space_mask (np.ndarray):            Mask in real space to apply background thresholding on the similarity array.
 
         """
-
         self.datacube = datacube
+        self.r_space_mask = r_space_mask
+        self.similarity = None
+        self.similarity_raw = None
+
+    def bg_thresholding(self, r_space_mask,):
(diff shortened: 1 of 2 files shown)


py4DSTEM/process/utils/cluster.py
```diff
@@ -15,21 +15,49 @@ class Cluster:
...
-        # Which neighbors to search
+        
+        """
+        Args:
+            q_space_mask : annular boolean q_space_mask to apply on the diffraction patterns
+            smooth_sigma : sigma for Gaussian smoothing of the diffraction patterns before calculating similarity
+            return_similarity : if True, return the similarity array
+        """
+        if self.r_space_mask is None:
+            self.set_mask(r_space_mask)
```
smribet: I don't think this function exists

py4DSTEM/process/utils/cluster.py
```diff
@@ -108,30 +146,27 @@ def get_color(self, cluster_index):
...
-
+        """
+        Args:
+            threshold : threshold for similarity to consider two pixels as part of the same cluster
+        """
+        
         sim_averaged = np.mean(self.similarity, axis=2)
 
+        # Assigning the background as 'counted' 
+        sim_averaged[~self.r_space_mask] = -1.0
```
smribet: This is causing errors. I would add checking earlier that your `r_space_mask` is a boolean array. If it is `None` you'll need to handle this earlier too.

smribet:
Hi @serinlee1065, thanks for submitting this! I adjusted the doc strings and deleted old code that was commented out. I removed the `bg_thresholding` function, as the same logic elsewhere in the code.  I think your real space mask needs to be edited. Let me know if you have any questions.

PR #733, opened 2025-10-22
DDF patch 11

maclariz:
I found out when processing large datasets that the DDFradialimage and DDFradialazimuthimage code is as slow as an old dog.  I then figured out that I had done the coding really stupidly, rebuilt it and it's now very fast.

Please approve this, along with the minor corrections to "show"


smribet (approved):
This is great! Thanks @maclariz

